PR #1, opened 2020-04-14
one test

joakimmoller:
`go test`

### DIFF
--- a/spf_test.go
+++ b/spf_test.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"testing"
+)
+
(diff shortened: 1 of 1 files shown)


local-minimum:
perhaps integrate with aws to run the tests too?

joakimmoller:
Do we have a build spec?

local-minimum (changes requested):
No, no buildspec yet, but as you intend to use github, I'd say this requires integration of testing on github, thus #2 merged and the yaml extended.

PR #2, opened 2020-04-14
Create go.yml

joakimmoller:
Trying out actions (standard template for go), didn't change anything from the proposed template.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,34 @@
+name: Go
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


local-minimum (commented):
if we are using github workflows then add testing too because those aren't run when building right?

joakimmoller:
yes, but there are no tests yet - it's in #1

local-minimum (approved):
OK, but not seeing the point of using github for this rather than aws unless we intend github to deploy the artifacts on aws?

PR #3, opened 2020-04-14
using internal

joakimmoller:
OK, finally got it to work, this includes #1. I closed #2 since we are aiming for AWS. This PR doesn't aim to set automation in place, Just to reorganise the code.

internal/innosatPUS - library to read ECSS packets
internal/cli - an example cli to read/view a rac-file

to build the cli `go build internal/cli`
to run tests `go test internal/innosatPUS`


.gitignore
```diff
@@ -13,3 +13,5 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+*.rac
```
local-minimum: Do we really want to exclude all rac files anywhere, could we not want them as test-fixtures?

internal/cli/cli.go
```diff
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"log"
 	"os"
+
+	pus "github.com/innosat-mats/rac-extract-payload/internal/innosatPUS"
 )
 
 func main() {
+	fmt.Println("hej")
```
local-minimum: remove

internal/cli/cli.go
```diff
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"log"
 	"os"
+
+	pus "github.com/innosat-mats/rac-extract-payload/internal/innosatPUS"
```
local-minimum: why different name between directory and package, and by having them in internal it signals that it's not suitable for others to use in their projects and isn't rather the opposite, that the "pus" part could be used with anyone that has a mats rac-data file at hand? 

I also don't like the name PUS or innosatPUS since innosat is just a company right? and PUS is just Packet Utilization Standard. Neither of which describes the contents. I think something like "payload" would be much more sensible.

local-minimum: Also how does this work before the code is at github?
And when you are working on a newer version of the package, locally, which version is found if the cli is run/build?
On your branch in VSC, I can't inspect definitions of the things in the pus package, neither renaming works. Have you done anything to the configuration, and is it possible to check in those setting?

internal/cli/cli.go
```diff
@@ -54,66 +57,66 @@ func main() {
 		}
 
 		// Data Field Header
-		dataHeader := TMDataFieldHeader{}
+		dataHeader := pus.TMDataFieldHeader{}
 		dhsize := binary.Size(dataHeader)
 		dhr := bytes.NewReader(buf[0:])
-		err = dataHeader.read(dhr)
+		err = dataHeader.Read(dhr)
```
local-minimum: can it not just be `reader` instead of another abbreviation (`dhr`) or just pass the reader directly as argument if it's not needed afterwards.

internal/cli/cli.go
```diff
@@ -54,66 +57,66 @@ func main() {
...
 					log.Fatal("stat", err)
 				}
 				fmt.Print(" STAT: ", stat)
 			}
 			if sid == 10 {
-				htr := HTR{}
-				err = htr.read(r)
+				htr := pus.HTR{}
+				err = htr.Read(r)
 				// if err != nil {
```
local-minimum: why is this commented out?

internal/cli/cli.go
```diff
@@ -54,66 +57,66 @@ func main() {
...
+				err = stat.Read(r)
 				if err != nil {
 					log.Fatal("stat", err)
 				}
 				fmt.Print(" STAT: ", stat)
 			}
 			if sid == 10 {
-				htr := HTR{}
-				err = htr.read(r)
+				htr := pus.HTR{}
```
local-minimum: This should be possible to refactor since the structure is identical to sid == 1, it should be possible to have a mapping of sid to parser.

local-minimum (changes requested):
It's close to impossible and extremely frustrating to review this code since the project started with a dump of unreviewed code into master. Perhaps we should just restart from scratch with no code, setup a testing and build system and then add small discrete PRs, one at a time?

internal/aez/ccd_power_regulators.go
```diff
@@ -25,6 +25,7 @@ type CPRU struct {
 	VOD3   uint16 // CCD3 Output Drain Voltage 0..4095
 }
 
-func (s *CPRU) read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, s)
+// Read CRPU
+func (r *CPRU) Read(buf io.Reader) error {
```
local-minimum:
```suggestion
func (cpru *CPRU) Read(buf io.Reader) error {
```

internal/aez/ccd_power_regulators.go
```diff
@@ -25,6 +25,7 @@ type CPRU struct {
...
+	return binary.Read(buf, binary.BigEndian, r)
```
local-minimum:
```suggestion
	return binary.Read(buf, binary.BigEndian, cpru)
```

internal/aez/general.go
```diff
@@ -21,6 +21,7 @@ type STAT struct { //(34 octets)
 	SPWEEP uint32 // SpaceWire received EEPs
 }
 
-func (s *STAT) read(buf io.Reader) error {
+// Read STAT
+func (s *STAT) Read(buf io.Reader) error {
```
local-minimum:
```suggestion
func (stat *STAT) Read(buf io.Reader) error {
```

internal/aez/general.go
```diff
@@ -21,6 +21,7 @@ type STAT struct { //(34 octets)
...
 	return binary.Read(buf, binary.BigEndian, s)
```
local-minimum:
```suggestion
	return binary.Read(buf, binary.BigEndian, stat)
```

internal/cli/cli.go
joakimmoller: yes

internal/aez/heater_regulators.go
```diff
@@ -33,6 +33,7 @@ type HTR struct {
 	HTR8OD uint16
 }
 
-func (s *HTR) read(buf io.Reader) error {
+// Read HTR
+func (s *HTR) Read(buf io.Reader) error {
```
local-minimum:
```suggestion
func (htr *HTR) Read(buf io.Reader) error {
```

internal/aez/heater_regulators.go
```diff
@@ -33,6 +33,7 @@ type HTR struct {
...
 	return binary.Read(buf, binary.BigEndian, s)
```
local-minimum:
```suggestion
	return binary.Read(buf, binary.BigEndian, htr)
```

internal/aez/power.go
```diff
@@ -18,6 +18,7 @@ type PWR struct {
 	PWRP3C3 uint16 // +3V3 current sense 0..4095
 }
 
-func (s *PWR) read(buf io.Reader) error {
+// Read PWR
+func (s *PWR) Read(buf io.Reader) error {
```
local-minimum:
```suggestion
func (pwr *PWR) Read(buf io.Reader) error {
```

internal/aez/power.go
```diff
@@ -18,6 +18,7 @@ type PWR struct {
...
 	return binary.Read(buf, binary.BigEndian, s)
```
local-minimum:
```suggestion
	return binary.Read(buf, binary.BigEndian, pwr)
```

internal/innosat/source.go
```diff
@@ -12,30 +12,37 @@ type SourcePacketHeader struct {
 	PacketLength          uint16
 }
 
-func (h *SourcePacketHeader) read(buf io.Reader) error {
+//Read SourcePacketHeader
+func (h *SourcePacketHeader) Read(buf io.Reader) error {
```
local-minimum: `h` may be OK here but I would prefer `sph`

local-minimum (commented):
Just some comments for now. In general the code looks good to me save some missing test coverage.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+GOTEST=$(GOCMD) test
+GOGET=$(GOCMD) get
+BIN_DIR=bin
+
+
+all: test build
+build: 
+		$(GOBUILD) -o $(BIN_DIR) -v ./...
+test: 
+		$(GOTEST) -v ./...
```
local-minimum, line 14: nice that `./...` worked here too.

internal/aez/ccd_power_regulators.go
```diff
@@ -25,6 +25,7 @@ type CPRU struct {
 	VOD3   uint16 // CCD3 Output Drain Voltage 0..4095
 }
 
-func (s *CPRU) read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, s)
+// Read CRPU
+func (cpru *CPRU) Read(buf io.Reader) error {
```
local-minimum, line 29: how about a test of a real CPRU buffer from a `.rac` file resulting in the expected CPRU struct?

internal/aez/general.go
```diff
@@ -21,6 +21,7 @@ type STAT struct { //(34 octets)
 	SPWEEP uint32 // SpaceWire received EEPs
 }
 
-func (s *STAT) read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, s)
+// Read STAT
+func (stat *STAT) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.BigEndian, stat)
```
local-minimum, line 26: how about a test of a real STAT buffer from a `.rac` file resulting in the expected STAT struct?

internal/aez/heater_regulators.go
```diff
@@ -33,6 +33,7 @@ type HTR struct {
 	HTR8OD uint16
 }
 
-func (s *HTR) read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, s)
+// Read HTR
+func (htr *HTR) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.BigEndian, htr)
```
local-minimum, line 38: how about a test of a real HTR buffer from a `.rac` file resulting in the expected HTR struct?

internal/aez/power.go
```diff
@@ -18,6 +18,7 @@ type PWR struct {
 	PWRP3C3 uint16 // +3V3 current sense 0..4095
 }
 
-func (s *PWR) read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, s)
+// Read PWR
+func (pwr *PWR) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.BigEndian, pwr)
```
local-minimum, line 23: how about a test of a real PWR buffer from a `.rac` file resulting in the expected PWR struct?

internal/innosat/source.go
```diff
@@ -0,0 +1,48 @@
...
+//SourcePacketHeader Source Packet Header
+type SourcePacketHeader struct {
+	PacketID              uint16
+	PacketSequenceControl uint16
+	PacketLength          uint16
+}
+
+//Read SourcePacketHeader
+func (sph *SourcePacketHeader) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.BigEndian, sph)
```
local-minimum, line 17: how about a test of a real SPH buffer from a `.rac` file resulting in the expected SPH struct?

internal/innosat/telecommand.go
```diff
@@ -0,0 +1,23 @@
...
+//TCDataFieldHeader (9 octets)
+type TCDataFieldHeader struct {
+	PUS            uint8
+	ServiceType    uint8
+	ServiceSubType uint8
+}
+
+// Read TCDataFieldHeader
+func (tcdfh *TCDataFieldHeader) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.BigEndian, &tcdfh)
```
local-minimum, line 17: how about a test of a real TC Data Field Header buffer from a `.rac` file resulting in the expected struct?

internal/innosat/telemetry.go
```diff
@@ -0,0 +1,30 @@
...
+	PUS             uint8
+	ServiceType     uint8
+	ServiceSubType  uint8
+	CUCTimeSeconds  uint32
+	CUCTimeFraction uint16
+}
+
+// Read TMDataFieldHeader
+func (tmdfh *TMDataFieldHeader) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.BigEndian, tmdfh)
```
local-minimum, line 19: how about a test of a real TM Data Field Header buffer from a `.rac` file resulting in the expected struct?

internal/innosat/telemetry.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// PUSVersion ...
+func (tmdfh *TMDataFieldHeader) PUSVersion() uint8 {
+	return (tmdfh.PUS << 1) >> 5
+}
+
+// Time ...
+func (tmdfh *TMDataFieldHeader) Time() uint32 {
```
local-minimum, line 28: I'll add test for this one in my PR so OK they are missing.

internal/ramses/ramses.go
```diff
@@ -0,0 +1,39 @@
...
+	Port   uint16
+	Type   uint8
+	Secure uint8
+	Time   uint32 // milliseconds since midnight
+	Date   int32  // days since 2000-01-01 00:00:00.00
+}
+
+// Read Ramses
+func (ramses *Ramses) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.LittleEndian, ramses)
```
local-minimum, line 22: how about a test of a real Ramses buffer from a `.rac` file resulting in the expected struct?

internal/ramses/secure.go
```diff
@@ -0,0 +1,21 @@
...
+	Port           uint16
+	Seq            uint16
+	Retransmission uint16
+	Ack            uint16
+	_              uint32
+}
+
+// Read fills a structure with data
+func (secure *Secure) Read(buf io.Reader) error {
+	return binary.Read(buf, binary.LittleEndian, secure)
```
local-minimum, line 20: how about a test of a real Secure buffer from a `.rac` file resulting in the expected struct?

local-minimum (changes requested):
I would also like the github CI from #2 modified to use the `make` command included in this PR and the protection of the master branch updated to require both build and test to pass.

joakimmoller:
I dont't think such tests add value to the code. Basically you say you want me to write tests to see if the a byte sequence e.g. 0xff 0xff actually is interpreted as a uint16 (65535) and that a set of byte sequences are resolved to the complete struct.

I'll add a github action  to run the tests.

local-minimum:
The tests would be there to ensure that the way the struct is described matches the specification and that the specification matches the actual data, which I think does add some value.

joakimmoller:
I disagree, the Read function can read any data, the methods on the structs are the ones that should validate the data. Still, reading bytes into a structure is not something we need to unit test that is already tested in the compiler (I suppose...)

An golang_example file  or an integration test on a rac record might be interesting.

local-minimum:
Yes sorry I didn't really mean as a unittest, but as an integration or systemtest

local-minimum (approved):
Just a question about the example system test, there's no risk of the test-code handling errors by the various `if err != nil` and thus hiding failings?

PR #11, opened 2020-04-16
Create go.yml

joakimmoller:


.github/workflows/go.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
+
+    - name: Test
+      run: go test -v ./...
```
local-minimum, line 34:
```suggestion
      run: go test -v ./...
      
    - name: Build
      run: go build -v ./...
```

local-minimum (approved):
maybe also add building so we ensure that works

PR #13, opened 2020-04-16
[WIP] Units

local-minimum:
To avoid a lot of code duplication this uses golang's "enums" and reflection to calculate the voltage of the CCD struct.


joakimmoller:
Looks good to me, but feels abit akward to give the fieldtype as parameter - but perhaps thats the way to do it. go for it!

PR #14, opened 2020-04-17
Power, Current, Resistance

### LINKED ISSUE #8: Convert to usable units

local-minimum:
This is a partial feature, it uses equations in AEZICD002 to translate various primary read/encoded data into physical units.

It was a bit vague in the case of the HTR struct what unit will be suitable for HTRxA and HTRxOD:

> Each regulator has two sensors, one used for regulation (HTRxA) and one for measuring the temperature
> (HTRxB). Along with these, the output drive value is also returned (HTRxOD).

So I left those as voltages, while I transformed HTRxB to resistance, which seems one step closer to temperature (if I've understood everything correctly).

For CPRUA/CPRUB (called CPRU), I ended at voltage as that is as far as equations in the spec take it.

For PWR I made PWRT into resistance, the other PWxV I left as voltage while the PWxC I left as currents as this seemed to be according to the specification.

I didn't do reports for any other struct yet (e.g. converting modes to strings)

Resolves #8 (will make explicit issue of producing temperature)

### DIFF
--- a/internal/aez/heater_regulators.go
+++ b/internal/aez/heater_regulators.go
@@ -33,7 +33,69 @@ type HTR struct {
 	HTR8OD uint16
 }
 
+//HTRReport housekeeping report returns data on all heater regulators in useful units.
+type HTRReport struct {
+	HTR1A  float64 // Heater 1 Temperature sense A voltage
+	HTR1B  float64 // Heater 1 Temperature sense B voltage
+	HTR1OD float64 // Heater 1 Output Drive setting voltage
+	HTR2A  float64
+	HTR2B  float64
+	HTR2OD float64
+	HTR3A  float64
+	HTR3B  float64
+	HTR3OD float64
+	HTR4A  float64
+	HTR4B  float64
+	HTR4OD float64
+	HTR5A  float64
+	HTR5B  float64
+	HTR5OD float64
+	HTR6A  float64
+	HTR6B  float64
+	HTR6OD float64
+	HTR7A  float64
+	HTR7B  float64
+	HTR7OD float64
+	HTR8A  float64
+	HTR8B  float64
+	HTR8OD float64
+}
+
 // Read HTR
 func (htr *HTR) Read(buf io.Reader) error {
 	return binary.Read(buf, binary.BigEndian, htr)
+}
+
+func htrVoltage(data uint16) float64 {
+	return voltageConstant * float64(data)
+}
+
+// Report returns a HTRReport with useful units
+func (htr *HTR) Report() HTRReport {
+	return HTRReport{
+		HTR1A:  htrVoltage(htr.HTR1A),
+		HTR1B:  htrVoltage(htr.HTR1B),
+		HTR1OD: htrVoltage(htr.HTR1OD),
+		HTR2A:  htrVoltage(htr.HTR2A),
+		HTR2B:  htrVoltage(htr.HTR2B),
+		HTR2OD: htrVoltage(htr.HTR2OD),
+		HTR3A:  htrVoltage(htr.HTR3A),
+		HTR3B:  htrVoltage(htr.HTR3B),
+		HTR3OD: htrVoltage(htr.HTR3OD),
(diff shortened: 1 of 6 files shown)


internal/aez/heater_regulators.go
```diff
@@ -5,35 +5,103 @@ import (
...
-	HTR6A  uint16
-	HTR6B  uint16
-	HTR6OD uint16
-	HTR7A  uint16
-	HTR7B  uint16
-	HTR7OD uint16
-	HTR8A  uint16
-	HTR8B  uint16
-	HTR8OD uint16
+	HTR1A  htr // Heater 1 Temperature sense A 0..4095
```
skymandr, line 20: sense should be sensor?

internal/aez/heater_regulators.go
```diff
@@ -5,35 +5,103 @@ import (
...
 )
 
+type htr uint16
+
+func (data htr) voltage() float64 {
+	return voltageConstant * float64(data)
+}
+
+func (data htr) resistance() float64 {
+	return 3.3*3900/data.voltage() - 3900
```
skymandr, line 15: Elsewhere you have spaces around the /.

internal/aez/power.go
```diff
@@ -5,20 +5,116 @@ import (
...
+}
+
+type pwrt pwr
+
+func (data pwrt) voltage() float64 {
+	return pwr(data).voltageADC()
+}
+
+func (data pwrt) resistance() float64 {
+	return 3300/data.voltage() - 1000
```
skymandr: Same here with /. Should really be a linter thing though.

skymandr: 3300 is really 3.3 * 1000, which to my mind looks clearer, since that is what the specs say. 

You could also perhaps make the base resistance an argument and use the same code for both NTC resistances, e.g.
`func (data pwrt) resistance() float64 { return resistance_ntc(1000) }`
and above
`func (data htrt) resistance() float64 { return resistance_ntc(3900) }`
but I have no idea if that is possible in `go`, or indeed if it would actually make anything much clearer since the code duplication is so small anyway.

internal/aez/heater_regulators.go
local-minimum: this is how it should be, formatting is automatic, it might be because of mix of multiplication and subtraction that spacing indicate grouping.

internal/aez/power.go
local-minimum: I considered such formats but since as you say the code duplication is small I wanted each formula as written in the spec in a single function, but I could do something about the `3.3*1000` (space probably won't be allowed)

joakimmoller: 3.3e3? the closes you can get to the spec. 🤓

skymandr: That would be acceptable. (EDIT: 3.3*1000 that is, 3.3e3 is not an improvement compared to the specs. See below.)

local-minimum: I did `3.3*1000` now and if anyone wants further changes they are free to make a pr :stuck_out_tongue_winking_eye:

skymandr: I think @local-minimum's solution is closer to the specs than 3.3e3. If you *really* want to get close to the specs, then go for 3.3*1e3, but 1000 is just as good to me.

joakimmoller (approved):
A sensor readout defined by the spec. (the actual measured sensor on the instrument) is what we have in the low level strucs i.e. STAT. Now each value in the low level struct is assigned to a type that describes the low level attributes (implicit by the type) . Wouldn't it be easier if all low level types implements an interface "Calibrated physical quantity" that have the "value" function that returns a float and "unit" function that returns  a string? Then we can fill a "Report" with instances of that interface type.

local-minimum:
Wouldn't it have to be that the types we have now are needed still but that they all satisfy the common interface? It sounds like something we would add ontop of this PR in that case? E.g. `od.value` calls `voltage` and unit returns `"V"`?

PR #15, opened 2020-04-17
Nicer interface for cli.go

### LINKED ISSUE #4: Introduce enums where suitable
* SourcePacketHeader.GroupingFlag, HeaderType, Type, APID
* TCDataHeader.ServiceType, ServiceSubType (at least move magic to the header file)

### LINKED ISSUE #5: Refactor aez servicetype 3, servicesubtype 25
Make an interface to avoid code duplication for the various sid cases.

local-minimum:
* Was a bit unsure maybe we should have names like `SID_PWR` but go doesn't seem to appreciate underscores.
* I updated `cli.go` so it makes sense with regards to TC and TM, which means it won't process anything in the racs since they don't seem to follow the specs. But I rather have our code make sense and then if one wants to use the cli while developing just comment  out the check or flip between the two.

Resolves #5
Resolves #4 (all that were in use at least, rest can be dealt with when we use them)

### DIFF
--- a/internal/innosat/source.go
+++ b/internal/innosat/source.go
@@ -12,8 +12,8 @@ type SourcePacketHeaderType uint
 //
 //TC is the source package type for telecommand
 const (
-	TM SourcePacketHeaderType = 0
-	TC SourcePacketHeaderType = 1
+	TC SourcePacketHeaderType = 0
+	TM SourcePacketHeaderType = 1
 )
 
 //SourcePacketHeader Source Packet Header
--- a/internal/innosat/source.go
+++ b/internal/innosat/source.go
@@ -5,6 +5,17 @@ import (
 	"io"
 )
 
+//SourcePacketHeaderType is the type of the source packet (TM/TC)
+type SourcePacketHeaderType uint
+
+//TM is the source package type for telemetry
+//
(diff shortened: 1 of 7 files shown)


internal/innosat/source.go
```diff
@@ -5,6 +5,17 @@ import (
...
 
+//SourcePacketHeaderType is the type of the source packet (TM/TC)
+type SourcePacketHeaderType uint
+
+//TM is the source package type for telemetry
+//
+//TC is the source package type for telecommand
+const (
+	TM SourcePacketHeaderType = 0
+	TC SourcePacketHeaderType = 1
```
joakimmoller:
```suggestion
	TC SourcePacketHeaderType = 0
	TM SourcePacketHeaderType = 1
```

internal/innosat/source_test.go
```diff
@@ -23,7 +23,7 @@ func TestType(t *testing.T) {
 		0}
 
 	ans := sph.Type()
-	if ans != 1 {
+	if ans != TC {
```
joakimmoller:
```suggestion
	if ans != TM {
```

joakimmoller (approved):

PR #22, opened 2020-04-17
updates to structures according to issue H

joakimmoller:


internal/aez/ccd_power_regulators.go
```diff
@@ -54,6 +57,7 @@ type CPRU struct {
 
 //CPRUReport structure
 type CPRUReport struct {
+	STAT
```
local-minimum: Do we really want the STAT as it is in the Report? It seems we want a STATReport here with processed readable data based on the stats info.

internal/aez/heater_regulators.go
```diff
@@ -23,18 +23,6 @@ type HTR struct {
 	HTR2A  htr
 	HTR2B  htr
 	HTR2OD htr
-	HTR3A  htr
```
local-minimum, line 26: wow :cry:  that would have been very nice not to have put in all that from the beginning.

internal/aez/ccd_power_regulators.go
```diff
@@ -34,6 +34,9 @@ func (data od) voltage() float64 {
 
 //CPRU structure
 type CPRU struct {
+	STAT uint8 // CPRU/CRB power status
```
local-minimum, line 37: Do we actually want this STAT still encoded or should there be a type here `STAT CPRUStat` where `CPRUStat` implements functions for reading out `OvervolatageFault(CCDSensor sensor) bool` and `Power(CCDSensor sensor) bool`. And `CCDSensor` is an 'enum' of masks for the sensors.

local-minimum (commented):
Also have you checked that no equation in the various conversions of encoded voltage to various units have been changed or not? If not, please make an issue that this needs to be checked.

internal/aez/ccd_power_regulators.go
joakimmoller: According to the section "changes" in the spec. document there was no algorithmic change (but haven't checked)

Didn't think of the interpretation of the STAT-field, this fix was mostly to be able to traverse the files (read the data into the structs) not to parse the info.

local-minimum (approved):
Maybe add an issue to parse the STAT field then? It kind of stands out as the only complex field we haven't parsed.

PR #26, opened 2020-04-21
CCD Image Data Pack

### LINKED ISSUE #10: Extract JPEG meta data from rac
Extract image meta data save next to jpeg (same filename?)

local-minimum:
Attempts to parse the CCD Image Data Pack according the the spec.

Contrary to the other `Read` functions this not only returns the error, but a fixed size uint16 array with the discarded columns data.

This means the read position should be right where one starts to read jpeg data.

I tried to include enough information from the spec so it should be possible to review, and please look carefully so I haven't made a mistake with all those bit operations.

Resolves #10


internal/aez/ccd_data.go
```diff
@@ -0,0 +1,170 @@
...
+//
+// FullTiming = 1, used even for pixels that are not read out
+const (
+	FasterTiming CCDGainTiming = iota
+	FullTiming
+)
+
+// Mode returns high/low signal mode
+func (gain CCDGain) Mode() CCDGainMode {
+	if gain&(1>>3) == 0 {
```
joakimmoller: (1>>3) :  shift a 1 three times to the right will always be zero. Maybe you want to change direction of the shift?
```
0b0000 00001 << 3 = 0b0000 1000
```

internal/aez/ccd_data.go
```diff
@@ -0,0 +1,169 @@
...
+type WDWMode uint8
+
+// WDWModeManual = 0, check WDWOV for overflows
+//
+// WDWModeAutomatic, window selected depending on input data
+const (
+	WDWModeManual    WDWMode = 0
+	WDWModeAutomatic WDWMode = 1
+)
+
```
joakimmoller: If you do like this VSC will help you with the constants (does not work with block comments):
```suggestion
const (
	// WDWModeManual = 0, check WDWOV for overflows
	WDWModeManual    WDWMode = 0
	// WDWModeAutomatic, window selected depending on input data
	WDWModeAutomatic WDWMode = 1
)

```

internal/aez/ccd_data.go
```diff
@@ -0,0 +1,169 @@
...
+// Wdw is the composite status of Image Window Mode
+type Wdw uint8
+
+// Mode returns the WDWMode type used encoded in Bit[7]
+func (wdw Wdw) Mode() WDWMode {
+	if (wdw & 0b0000001) == 0 {
+		return WDWModeManual
+	}
+	return WDWModeAutomatic
+}
```
joakimmoller: You have masked out bit 2, why not do the simpler
```suggestion
// Mode returns the WDWMode type used encoded in Bit[7]
func (wdw Wdw) Mode() WDWMode {
	return Wdw(wdw >> 7)
}
```

internal/aez/ccd_data.go
```diff
@@ -0,0 +1,169 @@
...
+
+// InputDataWindow returns which bits out of the depth of the CCD Images
+//
+// It ueses information in Bit[2..0]
+//
+// Returns major and minor bit of original CCD Image and error
+// if encoding not covered by specification.
+//
+// If full range of Bit[15..0] is used the JPEGQ should be 0xFF
+func (wdw Wdw) InputDataWindow() (int, int, error) {
```
joakimmoller, line 41: This function should return a function that does the bit operations used while reading the jpeg data.

internal/aez/ccd_data.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// FullTiming = 1, used even for pixels that are not read out
+const (
+	FasterTiming CCDGainTiming = iota
+	FullTiming
+)
+
+// Mode returns high/low signal mode, Bit[12]
+func (gain CCDGain) Mode() CCDGainMode {
+	if (gain & 0b00000000000001) == 0 {
```
joakimmoller: mask for bit 12: 0b00010000 00000000 or 0x1000

joakimmoller (changes requested):
Why don't you read the image data into a byte array when you define everything else?

internal/aez/ccd_data.go
local-minimum: mine seemed to work too, but nicer to just get info on the one being used rather than all of them I suppose.

internal/aez/ccd_data.go
local-minimum: For someone not too used with bit operations like me, a mask is immensely easier and clearer than bit-shifts. There's something intrinsic in them that hurts my brain. But as you kind of indirectly point out, is my mask really validating the 7th bit or the first?

internal/aez/ccd_data.go
local-minimum: Depends on how the data in the jpeg is encoded and if it is a reasonable path to assume we should manipulate the jpeg ourselves or use 3rd party tools for it. I think it's more reasonable to assume that we produce a jpeg as received and ask something else to convert it if we need to convert it.

internal/aez/ccd_data.go
skymandr: In my experience, long bit masks quickly get unreadable. Though it requires some getting used to, I believe hex-masks are considered better practice.

local-minimum: I think i could manage `f`,`2`,`1`, and `0` values for hexmasks but else you are asking me to make mistakes

skymandr: I'm asking you to learn not to make mistakes! ;)

skymandr: I certainly won't insist, but if I'm to work with similar problems in the code, I would prefer to use hex-based masks, for the same reason you prefer bit-masks. This comment is only an excuse to share this link, btw: https://flippybitandtheattackofthehexadecimalsfrombase16.com/
:-D

internal/aez/ccd_data.go
joakimmoller: You're masking out bit 0 the LSB.

An illustrated example on shift left (<<) https://www.computerhope.com/jargon/b/bit-shift.htm

internal/aez/ccd_data_test.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+}
+
+func TestCCDGain_Mode(t *testing.T) {
+	tests := []struct {
+		name string
+		gain CCDGain
+		want CCDGainMode
+	}{
+		{"Reads 12th bit", 0b0000000000001, LowSignalMode},
```
skymandr: isn't this the first bit?

local-minimum: yes I realized that it was wrong.

internal/aez/ccd_data.go
local-minimum: It's not really helping, it would help if we flipped the convention and wrote bit numbers left to right.

internal/aez/ccd_data.go
local-minimum: cool, but I kind of wonder if they learn bit to hex in a sense we would understand it or if they learn a pattern to enemy position equals reward.

joakimmoller (approved):

Issue #25, opened 2020-04-21
Make a CLI


local-minimum:
Create "CLI". 

Let the user select between jpeg or/and HK and select rac-files to process and output directory.

PR #27, opened 2020-04-21
Data flow from rac-file to sensor data

joakimmoller:
Controls the data flow from rac-file to sensor data. Reads ramses packages from file, converts to innosat packages with TM or TC content, and reads the Sensor data packages from the instrument.

Support for continuation packages (instrument level) is not implemented, but is accounted for. 

The design idea is to traverse through the packet layers and pull up the information back to the calling function. 

Supports #10  and #25


internal/common/extractdata.go
```diff
@@ -0,0 +1,49 @@
...
+	"log"
+	"sync"
+
+	"github.com/innosat-mats/rac-extract-payload/internal/aez"
+	"github.com/innosat-mats/rac-extract-payload/internal/innosat"
+	"github.com/innosat-mats/rac-extract-payload/internal/ramses"
+)
+
+// ExtractData reads Ramses data packages and extract the instrument data.
+func ExtractData(byteStream io.Reader, extract func(aez ...interface{}) (int, error)) {
```
local-minimum: I don't really like that `extract` has an interface that doesn't reflect the actual use just to satisfy `fmt.println`.
Construct an `InstrumentData` interface and add a small wrap around the print for this PR that has that interface.
I also don't like the name, the extraction is already done. To me this is the `callback` if anything.

internal/common/extractdata.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for ramsesPackage := range ramsesPackages {
+		select {
+		case err := <-ramsesErrors:
+			if err != nil {
+				log.Output(log.Llongfile, err.Error())
+				continue
+			}
+		default:
+			innosatPackage, err := innosat.DecodeSource(ramsesPackage.Payload)
```
local-minimum: Doesn't the innosat decoder have to keep information about what ramses packages it came from, and then the aez what innosat package and ramses it was in? Else we can't tag things with which rac-file they came from and such.

internal/ramses/reader.go
```diff
@@ -0,0 +1,58 @@
...
+// Package is complete Ramses package
+type Package struct {
+	Header       Ramses
+	SecureHeader Secure
+	Payload      []byte
+}
+
+//Packages reads Ramses packages from buffer
+func Packages(buf io.Reader, packageChannel chan<- Package, errorChannel chan<- error) {
+	defer close(packageChannel)
```
local-minimum: why is it this one closing the `chan`? What if there are more files coming?

internal/ramses/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	defer close(errorChannel)
+	var err error
```
local-minimum: why not just `err := header.Read(buf)` below?

internal/ramses/reader.go
```diff
@@ -0,0 +1,58 @@
...
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			errorChannel <- err
+			break
+		}
+
+		if !header.Valid() {
+			errorChannel <- errors.New("Not a valid RAC-file")
```
local-minimum: This isn't a very useful error, I would have liked to have a filename with it here, could we not require the streams be passed to in pair with the name/identifier cli or aws got for the file?

local-minimum: it should probably also be propagated further to the packageChannel and onto the aez chan.

internal/ramses/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	defer close(errorChannel)
+	var err error
+	for {
+		header := Ramses{}
+		err = header.Read(buf)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			errorChannel <- err
```
local-minimum: do we have test of this error handling? And also, as discussed on the next, wouldn't it be much more useful if error included the filename/key of the rac file?

internal/ramses/reader.go
```diff
@@ -0,0 +1,58 @@
...
+			break
+		}
+
+		if !header.Valid() {
+			errorChannel <- errors.New("Not a valid RAC-file")
+			break
+		}
+
+		secureHeader := Secure{}
+		if header.SecureTrans() {
```
local-minimum: this test, including the error handling seems not tested. And again, it seems more useful to have the filename/key, the alternative is ofc if the consumer of the error channel adds it.

local-minimum: test reading payload with secure header.

internal/innosat/reader.go
```diff
@@ -0,0 +1,58 @@
...
+
+// SourcePackage ...
+type SourcePackage struct {
+	Header      SourcePacketHeader
+	Payload     SourcePayload
+	Application []byte
+}
+
+// DecodeSource decodes a byte array to a SourcePackage
+func DecodeSource(data []byte) (SourcePackage, error) {
```
local-minimum: Consider my comments on ramses with regards to useful errors.

internal/innosat/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	header := SourcePacketHeader{}
+	err = header.Read(buf)
+	if err != nil {
+		return SourcePackage{}, err
+	}
+	if crc16.ChecksumCCITTFalse(data[:len(data)-2]) != binary.BigEndian.Uint16(data[len(data)-2:]) {
+		return SourcePackage{}, errors.New("checksum bad")
+	}
+	var payload SourcePayload
+	if header.Type() == TM {
```
local-minimum: looks like it should be a `switch`

internal/innosat/reader.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/howeyc/crc16"
+)
+
+// SourcePayload ...
+type SourcePayload interface {
+	PUSVersion() uint8
+}
+
+// SourcePackage ...
```
local-minimum: Since this one has the strange `Application` perhaps some info on what that is compared to `Payload`.

internal/innosat/reader.go
local-minimum: should we really `return SourcePackage{}, err` on problems and not `nil, err`?

internal/innosat/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	err = header.Read(buf)
+	if err != nil {
+		return SourcePackage{}, err
+	}
+	if crc16.ChecksumCCITTFalse(data[:len(data)-2]) != binary.BigEndian.Uint16(data[len(data)-2:]) {
+		return SourcePackage{}, errors.New("checksum bad")
+	}
+	var payload SourcePayload
+	if header.Type() == TM {
+		tmpayload := TMDataFieldHeader{}
```
local-minimum: camelCase this please

internal/innosat/reader.go
```diff
@@ -0,0 +1,58 @@
...
+		err = tmpayload.Read(buf)
+		if err != nil {
+			return SourcePackage{}, err
+		}
+		payload = tmpayload
+	}
+	if header.Type() == TC {
+		tcpayload := TCDataFieldHeader{}
```
local-minimum: camelCase please

internal/innosat/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	var payload SourcePayload
+	if header.Type() == TM {
+		tmpayload := TMDataFieldHeader{}
+		err = tmpayload.Read(buf)
+		if err != nil {
+			return SourcePackage{}, err
+		}
+		payload = tmpayload
+	}
+	if header.Type() == TC {
```
local-minimum: missing test for TC package

internal/innosat/source.go
```diff
@@ -34,7 +48,7 @@ func (sph *SourcePacketHeader) Version() uint {
 }
 
 // Type is either Telecommand or Telemetry
-func (sph *SourcePacketHeader) Type() SourcePacketHeaderType {
+func (sph SourcePacketHeader) Type() SourcePacketHeaderType {
```
local-minimum: why take a copy of the struct? Or am I missing something about the point of removing the `*`?

internal/innosat/source.go
```diff
@@ -43,18 +57,19 @@ func (sph *SourcePacketHeader) HeaderType() uint {
 	return uint((sph.PacketID << 4) >> 15)
 }
 
-func (sph *SourcePacketHeader) apid() uint16 {
+// APID ...
+func (sph SourcePacketHeader) APID() uint16 {
```
local-minimum: same

internal/innosat/source.go
```diff
@@ -43,18 +57,19 @@ func (sph *SourcePacketHeader) HeaderType() uint {
...
 // IsMainApplication says if packet is for main application
 func (sph *SourcePacketHeader) IsMainApplication() bool {
-	return sph.apid() == 100
+	return sph.APID() == 100
 }
 
 // GroupingFlags ...
-func (sph *SourcePacketHeader) GroupingFlags() uint {
-	return uint(sph.PacketSequenceControl >> 14)
+func (sph SourcePacketHeader) GroupingFlags() SourcePackageContinuationFlagType {
```
local-minimum: same

internal/innosat/source.go
```diff
@@ -12,8 +12,22 @@ type SourcePacketHeaderType uint
...
-	TC SourcePacketHeaderType = 0
-	TM SourcePacketHeaderType = 1
+	TM SourcePacketHeaderType = 0
+	TC SourcePacketHeaderType = 1
+)
+
+//SourcePackageContinuationFlagType type for continuation groups
+type SourcePackageContinuationFlagType uint
+
+//SPCont Continuation packet
```
local-minimum: I find I need an empty comment line for VSC not to just mangle the info as a single sentence with no spacing.

internal/innosat/telecommand.go
```diff
@@ -18,6 +18,6 @@ func (tcdfh *TCDataFieldHeader) Read(buf io.Reader) error {
 }
 
 //PUSVersion ...
-func (tcdfh *TCDataFieldHeader) PUSVersion() uint8 {
+func (tcdfh TCDataFieldHeader) PUSVersion() uint8 {
```
local-minimum: same

internal/innosat/telemetry.go
```diff
@@ -23,7 +23,7 @@ func (tmdfh *TMDataFieldHeader) Read(buf io.Reader) error {
 }
 
 // PUSVersion ...
-func (tmdfh *TMDataFieldHeader) PUSVersion() uint8 {
+func (tmdfh TMDataFieldHeader) PUSVersion() uint8 {
```
local-minimum: same

internal/innosat/source.go
local-minimum: also missing tests now that it is exported.

internal/aez/reader.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/innosat-mats/rac-extract-payload/internal/innosat"
+)
+
+//PackageType ...
+type PackageType interface {
+}
+
+// Packages ...
+func Packages(source <-chan innosat.SourcePackage, target chan<- PackageType) {
```
local-minimum: cant we replace the copypaste inside the inner switch with something like:
`data = readOrLog(structType interface{Read})` 

and if that is possible we could even have a `(sid)map[interface{Read}])` or something and remove the entire switch.

if not consider at least breaking apart into several functions, we not have a for -> if -> if -> switch

missing tests for:
* GroupingFlags not standalone
* TC header type
* Not main application
* HTR, PWR, CPRUA and CPRUB sids / packages. Both valid and bad versions.

local-minimum (changes requested):
I reviewed it top-down data flow-wise starting with cli and ending in aez, so probably useful to check my comments in that way since they refer to each other sometimes.

skymandr:
Have I understood correctly, if I assume that @joakimmoller has planned extensive changes to this, so that (re)viewing this is not a good idea at the current time?

local-minimum:
@skymandr yes I believe so

internal/common/extractdata.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// ExtractData reads Ramses data packages and extract the instrument data.
+func ExtractData(callback func(data DataRecord), streamBatch ...StreamBatch) {
+	var waitGroup sync.WaitGroup
+	records := make(chan DataRecord)
+	innosatPackages := make(chan DataRecord)
+	dataPackages := make(chan DataRecord)
+	go DataPackets(innosatPackages, dataPackages)
+	go Packets(records, streamBatch...)
```
local-minimum:
```suggestion
	go RamsesPackets(records, streamBatch...)
```

internal/common/extractdata.go
```diff
@@ -0,0 +1,41 @@
...
+
+	go func() {
+		waitGroup.Add(1)
+		for data := range dataPackages {
+			callback(data)
+		}
+		waitGroup.Done()
+	}()
+
+	for ramsesPackage := range records {
```
local-minimum: Add comment that this parses innosat packages.

internal/common/extractdata.go
```diff
@@ -0,0 +1,41 @@
+package common
+
+import (
+	"sync"
+)
+
+// ExtractData reads Ramses data packages and extract the instrument data.
+func ExtractData(callback func(data DataRecord), streamBatch ...StreamBatch) {
+	var waitGroup sync.WaitGroup
+	records := make(chan DataRecord)
```
local-minimum:
```suggestion
	ramsesPackages := make(chan DataRecord)
```

internal/common/extractdata.go
local-minimum:
```suggestion
	for dataRecord := range records {
```

internal/common/extractdata.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+		waitGroup.Done()
+	}()
+
+	for ramsesPackage := range records {
+		innosatPackage, err := DecodeSource(ramsesPackage.Buffer)
+		if err != nil {
+			ramsesPackage.SourceHeader = innosatPackage.Header
+			ramsesPackage.TMHeader = innosatPackage.Payload
+			ramsesPackage.Buffer = innosatPackage.ApplicationPayload
```
local-minimum: Remove?

internal/common/extractdata.go
```diff
@@ -0,0 +1,41 @@
...
+			ramsesPackage.Error = err
+			innosatPackages <- ramsesPackage
+			continue
```
local-minimum: Remove?

local-minimum (approved):

PR #34, opened 2020-04-27
A first real CLI

### LINKED ISSUE #25: Make a CLI

local-minimum:
# Interface
```
$ ./bin/rac 
Extracts information from Innosat-MATS rac-files

Usage: ./bin/rac [OPTIONS] rac-file ...

  -output string
    	Directory to place images and/or timeseries data
  -skip-images
    	Extract images from rac-files.
    	(Default: false)
  -skip-timeseries
    	Extract timeseries from rac-files.
    	(Default: false)
  -stdout
    	Output to standard out instead of to disk (only timeseries)
    	(Default: false)

No rac-files supplied
```

# Features
* Allows writing to standard out or disk (not really implemented yet).
* If disk, then requires `-output`
* If both `-skip-images` and `-skip-timeseries` then it just runs through silently without any output as ways to validate files

# Testing
It was difficult testing some aspects of this so:
* `main` is untested but kept minimal
* `getCallback` can't really test which type of callback is gotten nor how arguments were passed to the respective callback factories.
* some other holes, like what is usage printed and a bit vague test on the setting of processing time

Resolves #25

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ BIN_DIR=bin
 all: test build
 build:
 		$(GOBUILD) -o $(BIN_DIR) -v ./...
-		mv ${BIN_DIR}/cli ${BIN_DIR}/rac
 test:
 		$(GOTEST) -v ./...
 clean:
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,10 @@ BIN_DIR=bin
 
 
 all: test build
-build: 
+build:
 		$(GOBUILD) -o $(BIN_DIR) -v ./...
-test: 
+		mv ${BIN_DIR}/cli ${BIN_DIR}/rac
+test:
 		$(GOTEST) -v ./...
-clean: 
+clean:
 		$(GOCLEAN)
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -1,7 +1,10 @@
 package main
 
 import (
+	"errors"
+	"flag"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"time"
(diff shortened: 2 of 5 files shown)


internal/cli/cli.go
```diff
@@ -1,35 +1,126 @@
...
 	}
-	common.ExtractData(callback, batch...)
+	extractor(callback, batch...)
+	return nil
+}
+
+func init() {
+	skipImages = flag.Bool("skip-images", false, "Extract images from rac-files.\n(Default: false)")
+	skipTimeseries = flag.Bool("skip-timeseries", false, "Extract timeseries from rac-files.\n(Default: false)")
+	outputDirectory = flag.String("output", "", "Directory to place images and/or timeseries data")
```
skymandr, line 105: Does "" as default mean that it is mandatory? If so, then better as non-flag argument. Otherwise, make it clear what the default is ("./"?).

local-minimum: No it doesn't make it mandatory. And it's not, only mandatory if `-stdout` is not included. The default value ("") is just a hidden internal value to conform with interfaces of the `flag` lib and type restrictions and won't be accepted.

local-minimum: I find the UX better to have to explicitly state the output directory so one does not accidentally write loads of files into e.g. the directory one happened to execute from.

Makefile
```diff
@@ -8,9 +8,10 @@ BIN_DIR=bin
 
 
 all: test build
-build: 
+build:
 		$(GOBUILD) -o $(BIN_DIR) -v ./...
-test: 
+		mv ${BIN_DIR}/cli ${BIN_DIR}/rac
```
joakimmoller: Here we should use the simple `go install`, which installs the binary in the GO-path. Go clean can then remove the installation. The binary gets its name from the directory it is built in.
```
build: 
		$(GOBUILD)  ./...
install:
		$(GOINSTALL) ./...
clean:
		$(GOCLEAN) ./..
```

joakimmoller (approved):

Makefile
local-minimum: That's really annoying because I think `cli` is a bad name for the executable and a really good name of the directory in the code. Inversely I really don't like something like `rac` as a directory name in the code but it's a good name for the executable.

local-minimum:
```suggestion

```

PR #35, opened 2020-04-28
Csv writer

### LINKED ISSUE #21: Make a filesystem timeseries writer
Given header(s) and payload (PWRReport, HTRReport, CPRUReport (for CRPUA and CPRUB)) write content in a suitable transmission format to a file on disk.

local-minimum:
Introduces a proper csv writer factory that can write csv:s to one file per STAT/HTR/CPRU/PWR each per rac input file.

* Had to restructure the code in `common` to avoid circular imports
* Introduces interfaces for the communication between extractors and callbacks
* Adds CSV-prepared serialization of all structs that need it

# Testing

I think most of the code here managed to get tested. Note that there are a few custom written tests that aren't just auto-generated, even some of those that keep the outer structure of those generated have their internals rewritten.

I needed to fix the tests to cope with #36 but I'm out of energy in figuring and testing both when temperatures are inside and outside acceptable range.

Resolves #21


internal/aez/general.go
```diff
@@ -83,3 +104,41 @@ func (stat *STAT) Time(epoch time.Time) time.Time {
...
+func (stat STAT) CSVRow() []string {
+	var row []string
+	gpsTime := time.Date(1980, time.January, 6, 0, 0, 0, 0, time.UTC)
+	statTime := stat.Time(gpsTime)
+	row = append(row, statTime.Format(time.RFC3339Nano), fmt.Sprintf("%v", stat.Nanoseconds()))
+	val := reflect.Indirect(reflect.ValueOf(stat))
+	t := val.Type()
+	for i := 0; i < val.NumField(); i++ {
+		name := t.Field(i).Name
+		if name != "TS" && name != "TSS" {
```
BengtRydberg, line 132: vet inte vad dessa är (antar det är tiden), underlättar kanske med en kommentar

internal/aez/general_test.go
```diff
@@ -114,3 +114,221 @@ func TestSTAT_Nanoseconds(t *testing.T) {
...
+				MODE:    10,
+				EDACE:   11,
+				EDACCE:  12,
+				EDACN:   13,
+				SPWEOP:  14,
+				SPWEEP:  15,
+				ANOMALY: 16,
+			},
+			[]string{
+				"1980-01-06T00:00:08.000137329Z",
```
BengtRydberg, line 237: är det TSS 9 som generarar 000137329, bara så jag fattrar?

internal/cli/cli.go
```diff
@@ -115,12 +96,13 @@ func main() {
 		fmt.Println("\nNo rac-files supplied")
 		os.Exit(1)
 	}
-	callback, err := getCallback(*stdout, *outputDirectory, *skipImages, *skipTimeseries)
+	callback, teardown, err := getCallback(*stdout, *outputDirectory, *skipImages, *skipTimeseries)
```
BengtRydberg, line 99: it is a mixture of os.Exit, log.Fatal, and return err  in this function. Is this by purpose?

internal/aez/ccd_power_regulators.go
```diff
@@ -101,3 +103,30 @@ func (cpru *CPRU) Report() CPRUReport {
...
 }
+
+//CSVSpecifications returns the specs used in creating the struct
+func (cpru CPRU) CSVSpecifications() []string {
+	return []string{"AEZ", Specification}
+}
+
+//CSVHeaders returns the field names
+func (cpru CPRU) CSVHeaders() []string {
+	val := reflect.Indirect(reflect.ValueOf(cpru.Report()))
```
BengtRydberg: is is impossible to make a general CSVheaders function that you call from inside this and all similar funtions? so you do not need to "duplicate" code.

BengtRydberg (commented):
ser bra ut, men väldigt mycket kod.

internal/aez/general_test.go
local-minimum: Ja, borde tagit ett annat tal som t.ex. `0xc000` som ger 0.75

internal/aez/general.go
local-minimum: Yes det är TimeSeconds och TimeSubSeconds. Lägger in en kommentar.

internal/cli/cli.go
local-minimum: No not really, perhaps nice with `log.Fatal` in general

internal/aez/ccd_power_regulators.go
local-minimum: For just the Headers it might be quite possible, for the values it's less so... I'll check if I can figure something out.

internal/aez/csv_header.go
```diff
@@ -0,0 +1,24 @@
...
+		}
+	}
+	return false
+}
+
+func csvHeader(obj interface{}, exclude ...string) []string {
+	var fields []string
+	val := reflect.Indirect(reflect.ValueOf(obj))
+	t := val.Type()
+	for i, l := 0, val.NumField(); i < l; i++ {
```
BengtRydberg, line 18: this for loop looks strange. since l=0 it does not look like its doing any iterations . maybe I misunderstand how it works. Does it do the same as:

for i := 0; i < val.NumField(); i++ { ?

internal/aez/heater_regulators.go
```diff
@@ -140,3 +142,39 @@ func (htr *HTR) Report() HTRReport {
...
+	values := make([]string, val.NumField())
+	t := val.Type()
+	for i := range values {
+		valueField := val.Field(i)
+		if t.Field(i).Name == "WARNINGS" {
+			if valueField.Len() == 0 {
+				values[i] = ""
+			} else {
+				var errs = make([]string, valueField.Len())
+				for j, l := 0, valueField.Len(); j < l; j++ {
```
BengtRydberg, line 163: samma som ovan.

internal/aez/power.go
```diff
@@ -160,3 +162,38 @@ func (pwr *PWR) Report() PWRReport {
...
+	values := make([]string, val.NumField())
+	t := val.Type()
+	for i := range values {
+		valueField := val.Field(i)
+		if t.Field(i).Name == "WARNINGS" {
+			if valueField.Len() == 0 {
+				values[i] = ""
+			} else {
+				var errs = make([]string, valueField.Len())
+				for j, l := 0, valueField.Len(); j < l; j++ {
```
BengtRydberg, line 188: samma som ovan

BengtRydberg (commented):
looks good, except that I do not understand some for loops. can you explain them so I can approve?

internal/aez/csv_header.go
local-minimum: no
```
x, y := 1, 2
```
gives `x=1` and `y=2`

skymandr: Why is assigning `l` better than just `for i := 0; i < val.NumField(); i++`?

local-minimum: because it introspects to get the number of fields and introspection/reflection is slow so figured it be better this way.

local-minimum: feels like a fairly standard thing when you have to iterate over something of a flexible/unknown size in fast/typed languages to calculate the length and then iterate against that. I think this is very idiomatic go, but I could be wrong.

BengtRydberg: now I understand it, and I am fine with this, I misread how it works.

BengtRydberg (approved):

PR #37, opened 2020-04-29
12 bit jpeg grayscale to image.Gray16 decoder

joakimmoller:
Created a very first draft/proof of the possibilities to send jpeg-data to jpeglib to get the pixel values of the image. This solution uses CGO. I've provided a jpeglib with support for jpegs encoded with 12-bit color depth.

The implementation limited but sufficient, it supports only 12-bit greyscale images with known size. I've provided an example test to show the intended use. 

`make` will build and install (no tests) the binary in `$GOPATH/bin`


skymandr:
Quick comment: `example16gray.jpg` seems to be 8bit when I look at it:
`$ identify example16gray.jpg`
`example16gray.jpg JPEG 20x20 20x20+0+0 8-bit Grayscale Gray 256c 640B 0.000u 0:00.000`

joakimmoller:
corrected filename

Issue #20, opened 2020-04-17
Make a filesystem JPEG writer


local-minimum:
image data spans over multiple platform packages. Grouping of several packages necessary . Data is encoded in jpeg.

Needs #10

skymandr:
We think this is as easy as reading the correct binary blob and saving it with the correct filename.

local-minimum:
Me too, but need to handle file split over files though so needs to recognize when it should append and when it should start from the beginning.

joakimmoller:
Note that CCDImagePack should be read to the struct as LittleEndian to get the correct values.

Data comes in two different forms when JPEGQ>100 it is raw pixel data, other wise a jpeg file. Better to save everything as a png, need synchronization with #33 ?

PR #42, opened 2020-05-04
Aggregation of multi-packages and CCD Header parsing

local-minimum:
Part 1 of #20
* Adds aggregator for multipackets
* Parses transparency data and sends it to callback

## Testing 

Coverage of most parts but I don't really know if it is currently possible to reach all error paths so not those.

## Tip finding parsing errors
(best this is included in the `--help` by @joakimmoller  who probably already moved `cli.go`):
```
./bin/cli -stdout my.rac | grep -E -e".*Error:[^<:]+" -o
```
or if you want the `Buffer` conents which can be rather large if you are unlucky
```
./bin/cli -stdout my.rac | grep -E -e".*Error:[^<]+" -o
```


internal/aez/general.go
```diff
@@ -70,6 +70,29 @@ const (
 	PM RID = 30
 )
 
+func (rid RID) String() string {
```
joakimmoller, line 73: This file "general.go" was intended to be related to the "General status housekeeping structure" (STAT), RID is related to CCD data?

local-minimum: I think we should move STAT to a `stat.go` or at least I find it confusing finding it.

local-minimum: RID is the transparency data equivalent of SID and so yes it is related to CCD but not only.

internal/extractors/aez_reader.go
```diff
@@ -17,28 +19,117 @@ type PackageType interface {
...
+			// Panic if has started multipack
+			if multiPackBuffer != nil {
+				target <- prepareUnfinishedMultiPackPanic(multiPackBuffer, sourcePacket)
+				multiPackBuffer = nil
+			}
+			target <- processSourcePacket(reader, sourcePacket)
+		case innosat.SPStart:
+			// Panic if hasn't concluded previous
+			if multiPackBuffer != nil {
+				target <- prepareUnfinishedMultiPackPanic(multiPackBuffer, sourcePacket)
```
joakimmoller: You could use another channel as a queue, i.e. sending a part of a jpeg ([]byte) to a go routine that simply appends every chunk it gets to a larger byte array. Then the Exctract data could wait for the "result" if it recieves a "start header" package - giving the extract data function the resposibility to asseble the information header + data.

local-minimum: if we argue for that, I'd say this outer tier should not be an aez_reader at all but only be a package sorter so we have

```
sorter -> multipack aggregator
 |                      |
 \-------------------------------------> eaz_reader
```

but I don't really see how it does anything but adds complexity.

I really don't like the idea of having the first tier peaking into the packages to see what they are.

local-minimum: Or well a split of the `sorter`/`aggregator` and the `aez_reader` would make sense though. I'll do that.

local-minimum: I consider the aggregator I've now added to resolve this.

internal/extractors/aggregator.go
```diff
@@ -0,0 +1,111 @@
...
+			// Report standalone pack
+			target <- sourcePacket
+		case innosat.SPStart:
+			// Produce error for unfinished multipack lingering
+			if multiPackBuffer != nil {
+				target <- makeUnfinishedMultiPackError(multiPackBuffer, sourcePacket)
+			}
+
+			// Start new multipack
+			multiPackBuffer = bytes.NewBuffer([]byte{})
```
joakimmoller: You dont't need to reallocate memory every time, use Reset() to reuse the backing array (initiate the buffer once at startup). Check started multipack with MultiPackStart or another bool.

internal/extractors/aggregator.go
```diff
@@ -0,0 +1,111 @@
...
+			if multiPackBuffer == nil {
+				sourcePacket.Error = errors.New("got continuation packet without a start packet")
+				target <- sourcePacket
+				// Create a new one to capture remaining SPCont and attempt parse
+				multiPackBuffer = bytes.NewBuffer([]byte{})
+				multiPackStart = sourcePacket
+			}
+
+			// Concat SPCont packet
+			_, err := io.Copy(multiPackBuffer, reader)
```
joakimmoller: multiPackBuffer.ReadFrom(reader) ?

internal/aez/general.go
joakimmoller: neither of them (RID/SID) belongs in this file.

joakimmoller: OK to rename the file to stat.go but then, the files containing htr,pwr etc. also should be renamed.

internal/extractors/aggregator.go
local-minimum: isn't it pretty much equivalent?

internal/extractors/aggregator.go
local-minimum: fixed (multiPackStart is a struct so a bit uncomfortable checking against it)

internal/extractors/aggregator.go
local-minimum: (I changed though)

internal/aez/general.go
local-minimum: will make issue for renaming instead since it may be awkward right now that we are actively coding in the aez package.

internal/extractors/aez_reader.go
```diff
@@ -1,70 +1,54 @@
...
+		} else {
+			sourcePacket.Error = fmt.Errorf("%v | %v", exportableErr, bufErr)
+		}
+	} else {
+		sourcePacket.Error = exportableErr
+	}
+	sourcePacket.Data = exportable
+	sourcePacket.Buffer = buf
+	return sourcePacket
+}
```
joakimmoller: We are creating a byte array (buf) sized as the number of the remaining bytes in the bytes.Reader then extracting the contents out of the bytes.reader - what could go wrong?

What if the reader was a bytes.Buffer instead? (satifies io.reader)
Then you could do 
```sourcePacket.Buffer = reader.Bytes()  ```
That would remove the need for addData function?

**Review by joakimmoller — CHANGES_REQUESTED** [review]
Generally nice work! I don't the specific use of the tests in instrument_test but I don't consider them as harmful :)

There is one simplification I would like to see

File: internal/extractors/aez_reader_test.go [hunk 6]
```diff
@@ -51,8 +69,27 @@ func TestDecodeAEZ(t *testing.T) {
 				},
 			},
 			aez.STAT{},
+			aez.SIDSTAT,
+			aez.RID(0),
+			0,
 			true,
 		},
+		{
+			"Transparnt Data package",
```
skymandr:
```suggestion
			"Transparent Data package",
```

internal/extractors/aggregator.go
```diff
@@ -0,0 +1,113 @@
...
+				sourcePacketCopy := sourcePacket
+				sourcePacketCopy.Error = err
+				target <- sourcePacketCopy
+			}
+		case innosat.SPStop:
+			// Report error missing start pack
+			if !multiPackStarted {
+				// Report error
+				sourcePacketCopy := sourcePacket
+				sourcePacketCopy.Error = errors.New("got stop packed without a start packet")
```
skymandr:
```suggestion
				sourcePacketCopy.Error = errors.New("got stop packet without a start packet")
```

Also, why don't these get their own error functions?`No interesting data to trace, I guess?

internal/extractors/aggregator_test.go
```diff
@@ -0,0 +1,211 @@
...
+		multiPackBuffer *bytes.Buffer
+		sourcePacket    common.DataRecord
+	}
+	tests := []struct {
+		name string
+		args args
+		want common.DataRecord
+	}{
+		{
+			"Adds error an buffer to DataRecord",
```
skymandr:
```suggestion
			"Adds error and buffer to DataRecord",
```

skymandr:
A few typos and a question. That aside, I'm happy if @joakimmoller is happy.

internal/extractors/aez_reader.go
local-minimum: The main reason for the function was to not duplicate the assignment of `Error` `Data` and `Buffer` in the already a bit complex function where `addData` is used.

I don't know what could really go wrong with the reader, but it felt problematic not to handle errors from functions that return them. If I understand it correctly, not receiving errors from functions that produce them causes panic / stop of execution and that seemed like making the program brittle.

I was also considering buffers instead of readers, because readers allows seeking which is contrary to how we want to use them. It would make the copy-pasta if removing the function much less annoying.

local-minimum:
@joakimmoller I'd say that `instrument_test.go` is exactly one of the cases were it's extra important to fully test :wink: Because the implementation is a lot of copy paste of similar things it's much to easy to miss something or insert a typo and it can be hell to find such errors. By making all those tests I would have to have made the same mistake twice, which is much less probable.

joakimmoller:
take a look at #53 to see what I meant.

joakimmoller (approved):

PR #43, opened 2020-05-05
Raw data extraction from 12-bit jpegs

### LINKED ISSUE #33: Convert 12-bit jpegs to png
Go has its own jpeg encoder/decoder "image/jpeg" it is unclear if that supports 12-bit jpegs.

libjpeg from i ijg compiles nicely with 12-bit encoding/decoding and go can interface with c-libraries. In theory it would be possible decode the 12-bit jpeg via libjpeg to extract each pixel value and save it into a png-image.

joakimmoller:
This solution uses cgo as an interface to a provided version of jpeg that supports 12-bit color depth. The c code follows the decompression example pattern provided as instructions from jpeglib. It will crash ugly if the input cannot be parsed as a 12-bit grayscale jpeg image.

An integration test as an "Example" on a real MATS image is provided.


internal/decodejpeg/lib/pkgconfig/libjpeg.pc
```diff
@@ -0,0 +1,10 @@
+prefix=/home/joakim/dev-12bit
```
skymandr: This doesn't look intentional.

internal/decodejpeg/lib/libjpeg.la
```diff
@@ -0,0 +1,41 @@
...
+
+# Should we warn about portability when linking against -modules?
+shouldnotlink=no
+
+# Files to dlopen/dlpreopen
+dlopen=''
+dlpreopen=''
+
+# Directory that this library needs to be installed in:
+libdir='/home/joakim/dev-12bit/lib'
```
skymandr: Same here as below.

internal/decodejpeg/lib/pkgconfig/libjpeg.pc
joakimmoller: that is info about how it is built - I can remove the pkgconfig dir, but I thought it would be good info.

internal/decodejpeg/lib/libjpeg.la
joakimmoller: same here, we don't need the .la file - but I can't make it look any different.

internal/decodejpeg/lib/pkgconfig/libjpeg.pc
joakimmoller: OK, removed

internal/decodejpeg/lib/libjpeg.la
joakimmoller: OK, removed

internal/decodejpeg/decode.go
```diff
@@ -0,0 +1,36 @@
...
+// #include "decode.h"
+import "C"
+import (
+	"bytes"
+	"encoding/binary"
+	"unsafe"
+)
+
+//JpegImageData converts a grayscale image encoded in 12-bit jpeg to raw data
+func JpegImageData(jpegData []byte) ([]uint16, int, int, error) {
```
skymandr: Kanske vore trevligt med namngivna utargument för dokumentationens skull?

internal/decodejpeg/decode_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+	rreader := bytes.NewReader(ReferenceData)
+	refImg, err := pnm.Decode(rreader)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	fmt.Printf("Reference Data md5sum: %x\n", md5.Sum(refImg.(*image.Gray16).Pix))
+	//Output:
```
skymandr, line 42: Borde det inte vara en assert på detta?

joakimmoller: Output -kommentaren är asserten, exemplet failar om vi inte får samma resultat.

internal/decodejpeg/decode.go
joakimmoller: jepp, fixar det.

internal/decodejpeg/decode_test.go
skymandr: Ah, tufft. Och lite förvirrande.

skymandr (approved):
Looks good to me. We can handle the panic attacks as they occur.

PR #53, opened 2020-05-06
Suggestion

joakimmoller:
What I tried to say...


local-minimum:
I didn't see you did this, I think I did exactly the same thing just right now.

PR #56, opened 2020-05-07
Issue 28 interpret stat in cpru

### LINKED ISSUE #28: Interpret the STAT uint8 in the CPRU field
Issue H of the AEZ ICD added a field STAT to the CPRU(A/B) structure. Write interpretation logic for this. The field was added in #22 add it in the cvs output.

Expect a `[]bool` for overpowered and `[]bool` for powered to be gotten from the CPRU.STAT.

skymandr:
Interpret CPRU STAT field as bool arrays and format these for CSV.

Resolves #28


skymandr:
Question: is a single column each for `Overvoltage` and `Power` with a string of the format `true|true|false|true` good enough for the data to be properly usable when filtering future database searches..? This is human readable, but just saving the int might be easier when searching (can easily be filtered using bitmasks), whereas finding "rows where CCD2 had Overvoltage" is a bit difficult with a string. 

The most appealing alternative might be to have one column each for the 4 CCDs?

EDIT: Decided that I will make one column per CCD instead.

local-minimum (approved):
LGTM

PR #57, opened 2020-05-07
Leak fix in C

joakimmoller:
Very strange memory problem.


skymandr (approved):
This is most unsettling, but ok.

PR #59, opened 2020-05-07
Jpeg rowstride

### LINKED ISSUE #58: Memory allocation in "decode jpeg"
We have a strange issue with allocation of memory i the c-code that interacts with libjpeg. A workaround was introduced in PR #57. We currently don't know how much memory we need to allocate to hold a full image. In the first iteration we seemed to allocate too little which led to problems freeing memory!?

skymandr:
Revert @joakimmoller's quick fix and instead change chunk size in `memcpy`. It was set to `rowstride * BYTES_IN SAMPLE` but should - as I understand it - just be a single row, that is to say just `rowstride`. The fix is in the second commit. The first commit fixes annoyingly long lines in the `c` code.


joakimmoller (approved):
Very nice! This resolves #58

Issue #48, opened 2020-05-06
reorganize code in aez (and maybe more)


local-minimum:
especially `general.go` is confusing now.

Suggestion:
* Move everything STAT related from `general.go` to `stat.go`
* Rename the others to `htr.go`, `pwr.go`, `cpru.go`
* Not sure what to do with the two CCDs though.
* Remove bin to follow golang standards

local-minimum:
For extra credit the `Exportable` interface should be `Exporter` and the `ExportablePackage` fills no purpose and only causes a need to cast it into a `DataRecord` in the file-writer, so it should be removed. The methods that are on `DataRecord` should stay though, at least those that does something.

local-minimum:
`ExportablePackage` is no more.

PR #61, opened 2020-05-08
Writing of PNG images.

local-minimum:
This writes both raw pixels and compressed jpeg as png in the same output directory as the csvs using the pattern `nanoseconds.png`.

* Fixes so the RID/SID of continuation and stop packages are discarded in the aggregator
* Uses the PacketLength in the innosat reader.


internal/aez/ccd_data.go
```diff
@@ -114,6 +114,12 @@ func (gain CCDGain) Truncation() uint8 {
 	return uint8(gain & 0b1111)
 }
 
+// JPEGQUncompressed16bit is the value for non-12bit image data
+var JPEGQUncompressed16bit = uint8(101)
```
skymandr: why not `const`?

internal/exports/disk_writer.go
```diff
@@ -61,9 +57,51 @@ func DiskCallbackFactory(
...
+					shift,
+					imgFileName,
+				)
+
+				imgFile, err := os.Create(imgFileName)
+				if err != nil {
+					log.Printf("failed creating %s: %s", imgFileName, err)
+					panic(err.Error())
+				}
+				defer imgFile.Close()
```
skymandr, line 102: move to line 98?

skymandr:
```suggestion
				imgFile, err := os.Create(imgFileName)
				defer imgFile.Close()
				if err != nil {
					log.Printf("failed creating %s: %s", imgFileName, err)
					panic(err.Error())
				}

```
Or is that bad, i.e. will `Close` in itself likely cause an error?

internal/extractors/innosat_reader.go
```diff
@@ -16,6 +16,9 @@ type SourcePackage struct {
 	ApplicationPayload []byte
 }
 
+var crcChecksumLength int = 2
```
skymandr: Again, why not `const`? (I'm not sure what's best practice here.)

internal/aez/ccd_data.go
local-minimum:
```suggestion
const JPEGQUncompressed16bit = uint8(101)

```

local-minimum: I forgot about them :shrug:

internal/common/exportable.go
```diff
@@ -14,4 +14,6 @@ type ExportablePackage interface {
 	CSVSpecifications() []string
 	OriginName() string
 	AEZData() interface{}
+	RemainingBuffer() []byte
```
joakimmoller, line 17: This is a general comment on the interfaces:
1. I think the interfaces should be named after what they do, not what "data" it holds (be cause interfaces does not contain data it is just set of methods)
1. Exportable is a subset of Exportable package - could be dragged in...

```
type ExportablePackage interface {
    Exportable
	OriginName() string
    .
```

Better names could be somethinge like CVSExporter and CVSPackageExporter?

internal/extractors/innosat_reader.go
local-minimum:
```suggestion
const crcChecksumLength int = 2
```

internal/exports/disk_writer.go
local-minimum: no that should be fine

internal/aez/ccd_data.go
```diff
@@ -114,6 +114,13 @@ func (gain CCDGain) Truncation() uint8 {
 	return uint8(gain & 0b1111)
 }
 
+// JPEGQUncompressed16bit is the value for non-12bit image data
+const JPEGQUncompressed16bit = uint8(101)
+
+
+// NCOLStartOffset says how many more columns than reported the actual columns are
+var NCOLStartOffset uint16 = 1
```
skymandr:
```suggestion
const NCOLStartOffset uint16 = 1

```

internal/extractors/innosat_reader.go
```diff
@@ -16,6 +16,9 @@ type SourcePackage struct {
 	ApplicationPayload []byte
 }
 
+const crcChecksumLength int = 2
+var pusLengthOffset int = 1
```
skymandr:
```suggestion
const pusLengthOffset int = 1
```

internal/extractors/aggregator.go
```diff
@@ -16,13 +16,13 @@ func Aggregator(target chan<- common.DataRecord, source <-chan common.DataRecord
 	multiPackBuffer := bytes.NewBuffer([]byte{})
 	var multiPackStarted bool
 	var multiPackStart common.DataRecord
+	var sidRidLength = 2
```
skymandr:
```suggestion
	const sidRidLength = 2
```
maybe?

internal/exports/images.go
```diff
@@ -0,0 +1,99 @@
...
+			filename,
+			nPixels,
+			width,
+			height,
+			width*height,
+		)
+	}
+	buf := bytes.NewBuffer([]byte{})
+	if shift > 0 {
+		for idx := 0; idx < nPixels; idx++ {
```
joakimmoller: mutation of your own object then assigning back... 
```
for idx, pix := range pixels {
    pixels[idx] = pix << shift
}
```

internal/extractors/aggregator.go
joakimmoller: or binary.sizeof(uint16)

internal/extractors/innosat_reader.go
```diff
@@ -35,9 +41,11 @@ func DecodeSource(data []byte) (SourcePackage, error) {
 		return SourcePackage{}, err
 	}
 
+	sliceStart := binary.Size(header) + binary.Size(tmPayload)
+	sliceEnd := sliceStart + int(header.PacketLength) - binary.Size(tmPayload) - crcChecksumLength + pusLengthOffset
```
joakimmoller, line 45: This was not the error. The following line is better inline with the spec.
`sliceEnd := binary.Size(header) + int(header.PacketLength) - crcChecksumLength + pusLengthOffset`
even though
`sliceEnd := len(data) - crcChecksumLength `
still works.

internal/common/datarecord_test.go
```diff
@@ -359,3 +359,87 @@ func TestDataRecord_OriginName(t *testing.T) {
 		})
 	}
 }
+
+func TestDataRecord_RemainingBuffer(t *testing.T) {
```
joakimmoller, line 363: I think you should skip this test, you are adding 40 lines of boilerplate code for me to read for testing a getter of its simplest form.

internal/common/datarecord_test.go
```diff
@@ -359,3 +359,87 @@ func TestDataRecord_OriginName(t *testing.T) {
...
+				Buffer:       tt.fields.Buffer,
+			}
+			if got := record.RemainingBuffer(); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("DataRecord.RemainingBuffer() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestDataRecord_ParsingError(t *testing.T) {
```
joakimmoller, line 406: This test is for a direct getter - skip this.

joakimmoller:
In general nice and clean. I have some minor suggestions...

joakimmoller (approved):

internal/common/exportable.go
local-minimum: I appended it to #48 
When creating the interface I wasn't sure it it was going to be only for the csv exporting or also other formats but now that it seems to only need to cover our csv-needs `CSVExporter` might be better than `Exporter` (you can add that to the issue if you think so).

internal/common/datarecord_test.go
local-minimum: I'll leave it now to have time to merge this, and with the #48 , the getter would be removed anyhow.

internal/exports/images.go
```diff
@@ -0,0 +1,99 @@
...
+			pixels[idx] = pixels[idx] << shift
```
local-minimum:
```suggestion
		for idx, pix := range pixels {
			pixels[idx] = pix << shift

```

internal/extractors/innosat_reader.go
local-minimum: Not sure if you want me to change anything?

PR #64, opened 2020-05-08
Not ready yet to compile for windows.

joakimmoller:
Linking in the "linux static library" into the Windows binary didn't work (as suspected). Removing this feature for now - hopefully we can include a windows binary in later releases.


BengtRydberg (approved):
LGTM

PR #65, opened 2020-05-08
Win

joakimmoller:
The repo works fine on Windows 10 (64-bit) - one can build and test the code, had to update some tests to make them (more) platform independent. Cross compiling the binary for windows with mingw does not work on Ubuntu 18.04, it works on 20.04 though, but Focal is not yet available as a github action environment.


.github/workflows/go.yml
```diff
@@ -13,11 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.13
+    - name: Set up Go 1.14
       uses: actions/setup-go@v1
       with:
-        go-version: 1.13
+        go-version: 1.14
```
local-minimum, line 19: before merging this we should probably have a standup to inform everyone that they need to update their go, or can we keep it at 1.13 for now?

cmd/rac/cli_test.go
```diff
@@ -77,7 +77,7 @@ func Test_processFiles(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			dir, err := ioutil.TempDir("/tmp", "mats-testing")
+			dir, err := ioutil.TempDir("", "mats-testing*")
```
local-minimum: what's that asterisk?

local-minimum (commented):
I've only checked the go-code, and the actions so perhaps better someone checks the c-stuff

internal/decodejpeg/decode_windows.go
```diff
@@ -0,0 +1,36 @@
+package decodejpeg
```
skymandr: Is this code duplication necessary due to some cgo-magic? No way around it using `#IFDEF`s or similar in either `decode.go` or `decode.h`?

joakimmoller: Go uses convention over configuration - your linux go compiler will just ignore _windows files.

cmd/rac/cli_test.go
joakimmoller: adds a random number to the file (in the case of concurrent tests)

.github/workflows/go.yml
joakimmoller: don't we all use 1.14?

internal/decodejpeg/decode_windows.go
skymandr: It's still a lot of unnecessary duplication. It feels like this should be solvable in a better (_i.e._ more maintainable) way. In a c program, I would use _e.g._ `#ifdef _WIN64`, but for linking those directives would probably need to go in the makefile, in which case it might be tricky (and perhaps not go-like).

Bottom line: I don't like this, but I guess it might be ok for now. If we make a separate package of this, however, I think it should be cleaned up.

joakimmoller: OK, found a cleaner way to make it work.

.github/workflows/go.yml
local-minimum: Sorry, yes, I thought it was an upgrade of what we are running.

cmd/rac/cli_test.go
local-minimum: But I'm quite sure TempDir and TempFile did that without an asterisk. I have a feeling I logged the filenames and saw random numbers in the end.

skymandr (approved):
LGTM, though perhaps finish the discussion with @local-minimum before merging.

cmd/rac/cli_test.go
joakimmoller: yes - correct the asterisk is redundant at the end of the filname/dirname.

Issue #40, opened 2020-05-04
Parse OHBSE CCSDS TM Packet header?


local-minimum:
Remove the Secure Header from the code.

Is there useful info in the CCSDS TM header or just skip 16 bytes forward?

`Quality Indicator` and `Loss Flag`, maybe `Virtual Channel Frame Counter` could be of some use to diagnose any issues with data.

Use `uint8` anonymous padding for all other data in the header wihout any comments on what it is.

PR #66, opened 2020-05-11
Parse OHBSE CCSDS TM Packet Header

local-minimum:
This partially implements the OHBSE CCSDS TM Packet Header (I know that's a annoying name, but it's its name). It's part of the Ramses spec.

It implements `QualityIndicator`, `LossFlag` and `VCFrameCounter` and all three are exported as json and csv.

I have no good name for this struct and I admit it is quite horrible now, but we've also said that following naming in the spec is quite important.

## Testing

* Rewrote the stdout exporter test that was a annoying
* Added tests for the csv functions
* Updated existing tests

Relsolves #40


skymandr:
It feels like it is worth considering Orwell's advice here: ;)
>iv.  Break any of these rules sooner than say anything outright barbarous.


https://www.orwell.ru/library/essays/politics/english/e_polit

joakimmoller:
Since it is a header infront of all Telemetry packets, I suppose we can shorten it down to Ramses TM header

internal/common/datarecord.go
```diff
@@ -8,16 +8,16 @@ import (
...
-	SourceHeader innosat.SourcePacketHeader `json:"sourceHeader"` // Source header from the innosat platform
-	TMHeader     innosat.TMDataFieldHeader  `json:"tmHeader"`     // Data header information
-	SID          aez.SID                    // SID of the Data if any
-	RID          aez.RID                    // RID of Data if any
-	Data         Exportable                 `json:"data"`            // The data payload itself, HK report, jpeg image etc.
-	Error        error                      `json:"error,omitempty"` // First propagated error from the decoding process
-	Buffer       []byte                     `json:"-"`               // Currently unprocessed data (payload)
+	Origin                   OriginDescription          `json:"origin"`                   // Describes the origin of the data like filename or data batch name
+	RamsesHeader             ramses.Ramses              `json:"ramsesHeader"`             // Ramses header information
+	OhbseCcsdsTMPacketHeader ramses.OhbseCcsdsTMPacket  `json:"ohbseCcsdsTMPacketHeader"` // The CCSDS compliant OHBSE TM Packet header
```
joakimmoller: RamsesTMHeader?

internal/ramses/ohbse.go
```diff
@@ -0,0 +1,59 @@
+package ramses
```
joakimmoller: To me the filename is strange, OHB (the company) SE (the country), seems too generic for keeping the internal structure of one of the headers. "tmheader.go" would be good?

internal/extractors/ramses_reader.go
```diff
@@ -11,9 +11,9 @@ import (
 
 // Packet is complete Ramses packet
 type Packet struct {
-	Header       ramses.Ramses
-	SecureHeader ramses.Secure
-	Payload      []byte
+	Header             ramses.Ramses
+	OhbseCcsdsTMPacket ramses.OhbseCcsdsTMPacket
```
joakimmoller: TMHeader

joakimmoller (changes requested):

internal/ramses/ohbse.go
local-minimum: I know but as I understood it the header is the OHB (from Sweden)'s implementation of CCSDS standards for a TM header within the ramses specification. Kind of like when we wrote an extension to the RTZ in another project. "OHBSE CCSDS TM Packet" (header) is the actual name of it. My reservation for TM header is that it sounds like something that be part of / inside the telemetry packet, but as far as I undrestand this since it's part of the ramses spec, it's actually outside it.

local-minimum: Even more problematic is that in the ramses specification there are various "TM Packet" (it doesn't use the word header) so I think it would be problematic if we don't in some way explicitly say that its an OHBSE CCSDS TM Packet and a compromise would be to keep the file name as `ohbse.go` and call it `TMPacket` (omit the header entirely). CCSDS reference might not be important enough to be in either but just in documentation.

local-minimum: or we call the file `ohbse_ccsds_tm_packet.go` since files are not used in the code

joakimmoller (approved):
Yes!

Issue #67, opened 2020-05-11
Data race: concurrency problems


local-minimum:
Three re-runs with same settings in a row yields 3 different outputs / warning patterns. Could the packets be coming in random order to the aggregator? Or rather, that it's highly likely that they come in order but sometimes they don't? 

```
$ ~/go/bin/rac -output out racfiles/MyArchive1_20200428-204613_20200428-205254.rac
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:52 the TMHeader isn't recognized as either housekeeping or transparent data
Corrupt JPEG data: 2181 extraneous bytes before marker 0xd9
2020/05/11 09:09:53 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:09:53 the TMHeader isn't recognized as either housekeeping or transparent data
$ ~/go/bin/rac -output out racfiles/MyArchive1_20200428-204613_20200428-205254.rac
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
2020/05/11 09:10:11 the TMHeader isn't recognized as either housekeeping or transparent data
[...]

PR #69, opened 2020-05-11
Datarace temporary fix

joakimmoller:
issue #67 Describes a data race in the application. Here's a workaround until the real problem is solved.


local-minimum (approved):

skymandr:
Looks good to me too. If this is temporary, is there an issue for a less temporary fix?

PR #71, opened 2020-05-12
Sorted files

### LINKED ISSUE #31: Ensure sorted datastream as input to ExtractData
The raw ramses data packages to be processed need to be in order, otherwise we will lose images stored in two consecutive files.

Do this in `ramses_reader.go`

local-minimum:
Resolves #31 

First rac-record is read from each file, if valid it is used to sort input streams based on ramses time, then the remaining record are processed per file.


internal/common/remainingReader.go
```diff
@@ -0,0 +1,58 @@
...
+// Only supports files and bytes.Reader
+func NewRemainingReader(reader io.Reader) (*RemainingReader, error) {
+	var size int64 = -1
+	switch (reader).(type) {
+	case *os.File:
+		f, ok := reader.(*os.File)
+		if !ok {
+			return nil, errors.New("could not cast file")
+		}
+		fi, err := f.Stat()
```
skymandr: I think the convention for Swedish coders is to call it `fil` ;)

internal/extractors/extract_data_test.go
```diff
@@ -12,30 +12,30 @@ import (
...
 		0x31, 0x10, 0x03, 0x19, 0x00, 0x00, 0x12, 0x19, 0xe3, 0x39, 0x00, 0x01, 0x7f, 0x04, 0x02, 0x82,
 		0x04, 0x02, 0x02, 0x06, 0x01, 0x19, 0x12, 0x00, 0x00, 0x0c, 0xe3, 0x02, 0x00, 0x00, 0x00, 0x00,
 		0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x41, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 		0x00, 0x30, 0xfb,
-		//3 Too short buffer
-		0x90, 0xeb, 0x48, 0x00, 0x79, 0xd8, 0x00, 0x00, 0xd9, 0xb0, 0xc8, 0x04, 0xda,
-		0x1c, 0x00, 0x00, 0x00, 0x00, 0x2e, 0x02, 0x00, 0x00, 0x64, 0x00, 0x00, 0x00, 0x00, 0xcc, 0xcc,
+		//3 Too short buffer (innosat tm header length too short for a STAT) and also bad CRC
+		// Package length for ramses is correct (3rd & 4th byte here 0x46 = 70, and the 16 bytes header)
+		0x90, 0xeb, 0x46, 0x00, 0x79, 0xd8, 0x00, 0x00, 0xd9, 0xb0, 0xc8, 0x04, 0xda, 0x1c, 0x00, 0x00, 0x00, 0x00, 0x2e, 0x02, 0x00, 0x00, 0x64, 0x00, 0x00, 0x00, 0x00, 0xcc, 0xcc,
```
skymandr: why remove newline here..?

internal/extractors/ramses_reader.go
```diff
@@ -18,53 +19,101 @@ type Packet struct {
...
+
+	payload := make([]byte, header.Length)
+	n, _ := stream.Buf.Read(payload)
+	if n != int(header.Length) {
+		return common.DataRecord{
+			Origin:         stream.Origin,
+			RamsesHeader:   header,
+			RamsesTMHeader: tmHeader,
+			Error: fmt.Errorf(
+				"payload truncaded, only found %v bytes but needed %v (%v)",
```
skymandr, line 111: You break differently here than on line 89 (I prefer this style).

internal/extractors/ramses_reader_test.go
```diff
@@ -35,32 +43,236 @@ func TestDecodeRamses(t *testing.T) {
...
 				bytes.NewReader(
-					append([]byte{0x90, 0xeb, 17, 0}, make([]byte, 29+1)...))},
+					append([]byte{0x90, 0xeb, 17, 0}, make([]byte, minTotalSize-4+1)...),
+				),
+			},
 			[]byte{0},
 			false,
 		},
+		{
+			"a package with to short payload",
```
skymandr:
```suggestion
			"a package with too short payload",
```

skymandr (approved):
In general looks good. I think it would be best if @joakimmoller gives his explicit approval too, but I'll approve it none the less (they who dare win, etc...). There was one typo and a few places with inconsistent line breaks. You can safely ignore the comment about sour milk.

internal/extractors/extract_data_test.go
local-minimum: Oh that's not how I wanted it. The consistent way would have been to have 16 bytes for the first two rows since thats the length of the ramses and ramses tm headers.

internal/common/remainingReader.go
local-minimum: lol no this was how someone was calling on SO, so perhaps it's actually a go-convention.

internal/extractors/ramses_reader.go
local-minimum: I was a matter of length of line. I have no breaks on 89 because it's shorter.

joakimmoller (approved):

internal/extractors/extract_data_test.go
local-minimum: I made all 8 long and put comments showing Ramses Header,  Ramses TM Header, Ramses TM Payload.

internal/common/remainingReader.go
local-minimum: file was removed and used other solution after discussion with joakim

PR #72, opened 2020-05-12
Issue 44 handle TC verification

### LINKED ISSUE #44: Handle TC Verification
Should add convenience method(s) to TMHeader and get the output of TC Execution Sucess/Fail in a csv

skymandr:
Add handling of TC verification.

There is a lot of duplication, some of which could perhaps be avoided. One thing is omitted so far that I know, namely the error data field of the `TCExecFailure`. I wasn't sure if this was useful or perhaps proprietary. It is a variable length optional field "different for each APID and Failure Code", but I'm not sure where to find documentation for this.

Resolves #44


.gitignore
```diff
@@ -4,6 +4,7 @@
 *.dll
 *.so
 *.dylib
+.vscode
```
local-minimum, line 7: yay! been so close to commit it myself

internal/aez/verification_test.go
```diff
@@ -0,0 +1,350 @@
+package aez
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestTCAcceptSuccess_CSVSpecifications(t *testing.T) {
```
local-minimum, line 8: I've been thinking and perhaps we are a bit too eager to use these automated test structures when the entire body probably can be reduced to:

```go
tcv := TCAcceptSuccess{}
want := []string{"Specification", Specification}
if got := tcv.CSVSpecification(); !reflect.DeepEqual(got, want) {
  t.Errorf("TCAcceptSuccess.CSVSpecifications() = %v, want %v", got, want)
}
```

I think we only should use this automatic content as is when we have parameterized tests or other reasons it makes sense. They are still a good start in writing above code.

This applies to several tests.

internal/extractors/instruments.go
```diff
@@ -57,3 +57,29 @@ func instrumentTransparentData(rid aez.RID, buf io.Reader) (common.Exportable, e
 	}
 	return dataPackage, err
 }
+
+func instrumentVerification(subtype uint8, buf io.Reader) (common.Exportable, error) {
+	var dataPackage common.Exportable
+	var err error
+	switch subtype {
+	case 1: // TC Acceptance Report - Success
```
local-minimum: I think I'd prefer these as const of a dedicated type defined in `aez/verification.go` like we do for SID and RID

internal/innosat/telemetry.go
```diff
@@ -51,6 +51,15 @@ func (header *TMHeader) IsTransparentData() bool {
 	return header.ServiceType == 128 && header.ServiceSubType == 25
 }
 
+// IsTCVerification returns true if payload contains TC verification data
+func (header *TMHeader) IsTCVerification() bool {
+	return header.ServiceType == TelecommandVerification &&
+		(header.ServiceSubType == 1 ||
```
local-minimum: And use the same consts here, because now you have the magic numbers in two places and thus I feel even stronger about it.

local-minimum (changes requested):

internal/aez/verification_test.go
```diff
@@ -0,0 +1,102 @@
...
+	tcv := TCAcceptSuccessData{}
+	want := []string{"AEZ", Specification}
+	if got := tcv.CSVSpecifications(); !reflect.DeepEqual(got, want) {
+		t.Errorf("TCAcceptSuccessData.CSVSpecifications() = %v, want %v", got, want)
+	}
+}
+
+func TestTCAcceptFailure_CSVSpecifications(t *testing.T) {
```
local-minimum, line 16: I like these short tests!

internal/extractors/extract_data_test.go
```diff
@@ -68,13 +68,13 @@ func Example() {
 	// CCD3  got stop packet without a start packet
 	// Unknown RID: 303  unhandled RID Unknown RID: 303
 	//  STAT <nil>
-	//   the TMHeader isn't recognized as either housekeeping or transparent data
-	//   the TMHeader isn't recognized as either housekeeping or transparent data
+	//   the TMHeader isn't recognized as either housekeeping, transparent or verification data
```
local-minimum: FYI, this was really not a proper output here and my PR should fix it. It's just that errors that originated at a earlier stage were ignored and overwritten.

local-minimum (approved):
lgtm

PR #73, opened 2020-05-13
CCD metadata as CSV Output

local-minimum:
Since the CCDImagePackData has several complex types I figured it be more complex and confusing to use standard reflection to get contents out so instead I opted for hardcoding it. The order follows order of data in struct and for complex fields such as WDW the headers become "WDW Mode" and "WDW InputDataWindow"

I opted to have "WDW InputDataWindow" as a bit-range description like "15..2" which follows the pattern of how it is documented in the specification. @joakimmoller  That it's written "Bit[15..2]" rather than "Bit[2..15]" in the spec, could that imply we may need to inverse bits?.

We also have the mysterious "TIMING1" - "TIMING3" that we perhaps could refine further, but we need input from Chalmers if there's a need for it.

It uses GPS Time as epoch.

Partial sample output:
Screenshot from 2020-05-13 11-36-11

No BC (Bad Coulmns) will appear as `[]` and if you have several it will be `[1 33 515]`


internal/aez/ccd_data.go
```diff
@@ -155,21 +188,92 @@ type CCDImagePackData struct {
...
+		"TIMING3",
+		"NBC",
+	}
+}
+
+// CSVRow returns the exportable field values
+func (ccd CCDImagePackData) CSVRow() []string {
+	wdwhigh, wdwlow, _ := ccd.WDW.InputDataWindow()
+	return []string{
+		strconv.Itoa(int(ccd.CCDSEL)),
```
joakimmoller, line 249: I'd `fmt.sprintf("%i", ccd.CCDSL)` type of conversions

joakimmoller, line 249: I'd prefer `fmt.sprintf("%i", ccd.CCDSL)` type of conversions...

joakimmoller (approved):

internal/aez/ccd_data.go
local-minimum: I think `strconv` is preferred idiomatic way to convert ints and floats to strings and should be faster too.

PR #74, opened 2020-05-13
modified build triggering

### LINKED ISSUE #38: Something is wrong with github CI
I get an error: 
```
##[error]        heater_regulators_test.go:163: HTR.CSVHeaders() = [HTR1A HTR1B HTR1OD HTR2A HTR2B HTR2OD HTR7A HTR7B HTR7OD HTR8A HTR8B HTR8OD WARNINGS], want [HTR1A HTR1B HTR1OD HTR2A HTR2B HTR2OD HTR7A HTR7B HTR7OD HTR8A HTR8B HTR8OD]
```
on my branch on github, but not on my machine. The error indicates that the tests were actually not run on my branch as it is but rather on how my branch would be when I update against master. This is a rather confusing and annoying behaviour.

BengtRydberg:
Intention is that this should happen on push to any branch. And that no default merging
with master head is done. It is a bit difficult to test this. 
The checkout is discussed here: https://github.com/actions/checkout/issues/27

Resolves #38

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -20,9 +20,9 @@ jobs:
       run : sudo apt-get install -y gcc-mingw-w64
 
     - name: Check out code into the Go module directory
-       uses: actions/checkout@v1
-       with:
-         ref: ${{ github.head_ref }} 
+      uses: actions/checkout@v1
+      with:
+        ref: ${{ github.head_ref }} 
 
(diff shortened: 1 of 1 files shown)


local-minimum (approved):
LGTM, and perhaps try a fake branch after that would work on its own but not against master? or if possible to see on the action log, to make sure it works as we wish

PR #75, opened 2020-05-13
Performance

joakimmoller:
Trace analysis showed that png-encoding was the major event, blocking other threads to spawn. This runs the png file creation in a separate go routine.

### DIFF
--- a/internal/exports/disk_writer.go
+++ b/internal/exports/disk_writer.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/innosat-mats/rac-extract-payload/internal/aez"
 	"github.com/innosat-mats/rac-extract-payload/internal/common"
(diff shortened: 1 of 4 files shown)


internal/exports/disk_writer.go
```diff
@@ -241,6 +246,7 @@ func DiskCallbackFactory(
 		if tcvOut != nil {
 			tcvOut.close()
 		}
+		wg.Wait()
```
local-minimum, line 249: Is this the correct/optimal placement, can't we wait inside `teardown` function instead?

local-minimum (changes requested):

internal/exports/disk_writer.go
joakimmoller: ❓ It's already there!

local-minimum: haha I missed it. great!

local-minimum (approved):

PR #77, opened 2020-05-13
adding a dummy whitespace

BengtRydberg:
just a test


local-minimum:
Suggestion: making some small changes to Readme.md instead, It would be nice with a proper title and some smaller about.

PR #82, opened 2020-05-14
Moving stat to stat.go and cleaning up some tests

### LINKED ISSUE #48: reorganize code in aez (and maybe more)

local-minimum:
* Renamed `Exportable` to `Exporter`
* Now the `general.go` means general like it does in `innosat` and `ramses` packages. 

I'm not convinced that we gain much by renaming `power.go` to `pwr.go` so I'm just doing this bit and hope you agree that with it #48 can be considered resolved

Resolves #48


skymandr (approved):
LGTM

Issue #19, opened 2020-04-17
Make AWS S3 callback


local-minimum:
Like our diskwriter but writes all files to AWS instead.

PR #88, opened 2020-05-14
Images and their jsons to AWS S3

local-minimum:
* Adds ability to prefix with a project name (renaming of previous "output")
* Includes option to upload a project description file that will be named `ABOUT` on S3

Bucket top level (test is for us to abuse):
Screenshot from 2020-05-14 16-37-48

Contents in (test):
Screenshot from 2020-05-14 16-34-01

This starts the work on #19, next step is to also allow uploading of timeseries data.

# Testing

I managed to test the AWS callback factory, even though it's a quite special test. However I didn't see any reasonable way to test the actual uploader short of adding system tests but that would require credentials on github and all sorts of things.

I also didn't see any reasonable unittest for the `CCDImage.Image()` function. Or I could make one checking that the filename is as expected... if that feels valid enough here.


cmd/rac/cli.go
```diff
@@ -51,7 +51,7 @@ func getCallback(
 ) (common.Callback, common.CallbackTeardown, error) {
 	if project == "" && !toStdout {
 		flag.Usage()
-		fmt.Println("\nExpected an output directory")
+		fmt.Println("\nExpected an project")
```
skymandr:
```suggestion
		fmt.Println("\nExpected a project")
```

internal/exports/aws_s3_writer_test.go
```diff
@@ -0,0 +1,227 @@
...
+						NCOL:  1,
+						NROW:  2,
+					},
+				},
+				Buffer: make([]byte, 2*2*2), // 2x2 pixels, 2 bytes per pix
+			},
+			map[string]int{},
+		},
+		{
+			"Doesn't uploads errors",
```
skymandr:
```suggestion
			"Doesn't upload errors",
```

internal/exports/aws_s3_writer_test.go
```diff
@@ -0,0 +1,227 @@
...
+				},
+				Buffer: make([]byte, 2*2*2), // 2x2 pixels, 2 bytes per pix
+			},
+			map[string]int{
+				filepath.Join("myproj", "MyRac_5000000000.png"):  76,  // 8 + header
+				filepath.Join("myproj", "MyRac_5000000000.json"): 853, // length of the json
+			},
+		},
+		{
+			"Doesn't uploads image when told not to",
```
skymandr:
```suggestion
			"Doesn't upload image when told not to",
```

skymandr:
This looks good to me, but I don't feel comfortable enough with AWS to sign off on it. I suggest you let @joakimmoller have that honour.

local-minimum:
@skymandr  He's already on it :smile:

cmd/rac/cli.go
```diff
@@ -95,8 +109,10 @@ func processFiles(
 func init() {
 	skipImages = flag.Bool("skip-images", false, "Extract images from rac-files.\n(Default: false)")
```
joakimmoller, line 110: You are often using the bools with double negation "Don't skip images" - perhaps change the bool to be positive "images" true or false. This means the user still says -skip-images but we are using the negation.

internal/exports/aws_s3_writer_test.go
```diff
@@ -0,0 +1,227 @@
...
+			"Uploads description file",
+			args{
+				project:             "myproj",
+				descriptionFileName: "myfile.txt",
+				descriptionFileBody: []byte("Hello"),
+			},
+			common.DataRecord{},
+			map[string]int{
+				filepath.Join("myproj", "ABOUT.txt"): 5,
+			},
```
joakimmoller: This test fails on windows it is not supposed to be a file on disk? "myproj/ABOUT.txt" or filepath.ToSlash(filepath.Join("myproj", "ABOUT.txt")) works fine.

internal/exports/aws_s3_writer_test.go
```diff
@@ -0,0 +1,227 @@
...
+						EXPTS: 5,
+						JPEGQ: aez.JPEGQUncompressed16bit,
+						NCOL:  1,
+						NROW:  2,
+					},
+				},
+				Buffer: make([]byte, 2*2*2), // 2x2 pixels, 2 bytes per pix
+			},
+			map[string]int{
+				filepath.Join("myproj", "ABOUT.json"):            7,
```
joakimmoller: First line key fail on windows, but no the others. same as above.

internal/exports/disk_writer.go
```diff
@@ -82,21 +81,7 @@ func DiskCallbackFactory(
...
-				)
-				_, shift, _ := ccdImage.PackData.WDW.InputDataWindow()
-				img := getGrayscaleImage(
-					imgData,
-					int(ccdImage.PackData.NCOL+aez.NCOLStartOffset),
-					int(ccdImage.PackData.NROW),
-					shift,
-					imgFileName,
-				)
+				img, imgFileName := ccdImage.Image(pkg.Buffer, output, pkg.Origin.Name)
```
joakimmoller: Move this line into the go function below.

internal/exports/aws_s3_writer.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		if writeImages {
+			switch pkg.Data.(type) {
+			case aez.CCDImage:
+				ccdImage, ok := pkg.Data.(aez.CCDImage)
+				if !ok {
+					log.Print("Could not understand packet as CCDImage, this should be impossible.")
+					break
+				}
+				img, imgFileName := ccdImage.Image(pkg.Buffer, project, pkg.Origin.Name)
```
joakimmoller: this should go inside the go function below.

internal/exports/aws_s3_writer.go
```diff
@@ -0,0 +1,106 @@
...
+
+				wg.Add(1)
+				go func() {
+					pngBuffer := bytes.NewBuffer([]byte{})
+					png.Encode(pngBuffer, img)
+					upload(uploader, imgFileName, pngBuffer)
```
joakimmoller, line 90: The aws_s3_writer is very similar to disk_writer is it possible to generalize the control flow of saving to storage? upload() or "save to file" seems to be the discrimintor between the two callbacks?

joakimmoller (changes requested):
There is quite a lot of similar code in disk_writer.go and aws_s3_writer.go - It would be nice to se the control flow for prepareing the data for sending to storage sparate from the actual "upload" or "write to disk".

internal/exports/aws_s3_writer_test.go
local-minimum: if we are to support windows we really need github actions to run on both windows and linux in the near future and if we're running the tests on both, the prehaps there's little point in crosscompiling?

internal/exports/disk_writer.go
local-minimum: then the whole json bit needs to move too, which isn't bad i suppose.

internal/exports/aws_s3_writer.go
local-minimum: I was considering this, but I don't think it's the proper time for it. There's an inherent risk of over engineering it and at least the time-series data should also have been handled by the aws writer.

internal/exports/aws_s3_writer.go
local-minimum: done

internal/exports/aws_s3_writer_test.go
local-minimum: I fixed the test

internal/exports/aws_s3_writer_test.go
local-minimum: done

joakimmoller (approved):

PR #89, opened 2020-05-15
Buffered channels

joakimmoller:
Using buffered channels since I noticed that the go routings spends a lot of time waiting for synchronization between channels. The value 1000 is a educated guess since I know one picture is about ~2000 rac-records. Innosat channel seems to be the limiting channel.


local-minimum (approved):
LGTM, is the size set based on experimenting / random / recommendation? I guess it will set the upper limit of memory usage that may be needed, any risk this might get too high?

internal/extractors/extract_data.go
```diff
@@ -9,13 +9,15 @@ import (
 // ExtractFunction is the type of the ExtractData function
 type ExtractFunction func(callback common.Callback, streamBatch ...StreamBatch)
 
+const channelBufferSize int = 1000
```
skymandr: Nit-picking here, but are re there any cache chunk related reasons to think that 1024 might be better?

joakimmoller: OK, it looks better :)

joakimmoller:
@local-minimum: experimenting and guessing - yes, the application will use more heap - we're useing about 120Mb on my laptop (8cpus)

joakimmoller:
Sorry guys I added one more optimization moved the image conversion into the go routine for encoding png.

local-minimum (changes requested):
please undo these extra changes. They will directly conflict with my PR and the optimization here seems much lower priority to me.

internal/exports/disk_writer.go
```diff
@@ -83,7 +83,6 @@ func DiskCallbackFactory(
 					break
 				}
 				imgFileName := getGrayscaleImageName(output, pkg.Origin.Name, ccdImage.PackData)
```
local-minimum, line 85:
```suggestion
				imgFileName := getGrayscaleImageName(output, pkg.Origin.Name, ccdImage.PackData)


```

internal/exports/disk_writer.go
```diff
@@ -83,7 +83,6 @@ func DiskCallbackFactory(
 					break
 				}
 				imgFileName := getGrayscaleImageName(output, pkg.Origin.Name, ccdImage.PackData)
-
 				imgData := getImageData(
```
local-minimum:
```suggestion

				imgData := getImageData(
```

local-minimum: to keep file untouched.

local-minimum (approved):

PR #94, opened 2020-05-18
Linting, CI and Makefile

joakimmoller:
Added compilation directives to Martin's original work on CI. Moved dependicies to "third_party"


Makefile
```diff
@@ -1,18 +1,23 @@
...
-GOTEST=$(GOCMD) test
-GOGET=$(GOCMD) get
-GOINSTALL=$(GOCMD) install
+GOPATH ?= $(HOME)/go
+GOCMD := go
+GOBUILD := $(GOCMD) build
+GOCLEAN := $(GOCMD) clean
+GOTEST := $(GOCMD) test
+GOVET := ${GOCMD} vet
+GOLINT := golint
```
local-minimum: this breaks my env:
```
martin@mayak:~/go/src/github.com/innosat-mats/rac-extract-payload$ make lint
go vet ./...
golint ./...
make: golint: Command not found
Makefile:15: recipe for target 'lint' failed
make: *** [lint] Error 127
```

I have golint installed and previous solution did work. Why doesn't it work for you?

local-minimum (changes requested):
Also perhaps @skymandr should look at the changes to jpeg building since that's quite far outside my expertise

skymandr (approved):
Looks good to me. Two things come to mind, in case we find in the future that jpeg-handling is still a bottle neck: we could try adding `-O3` to the ccflags to ask the compiler to (aggressive) optimize the code, and we could try libjpegturbo, which is supposed to be faster than IJG:s implementation (and easier to get to work with 12-bit). Since premature optimization is the mother of a lot of evil, I'm not suggesting it now, but both may be worth keeping in mind, in case we find a need to chase cycles in the future.

skymandr:
(but please finish your discussion with @local-minimum before merging!)

local-minimum (approved):

PR #95, opened 2020-05-19
Generalize CSV writing

local-minimum:
Because we need a common interface for writing to disk and AWS this is a refactor of existing code.
It moves the bulk timeseries logic out from the callback and removes a lot of almost identical code blocks.

It doesn't yet introduce AWS writing, but if the underlying io.writer is of a type that uploads on close it should fit neatly into this pattern. And that's what I plan to add next.

I also uses interface with pointers to structs :man_dancing: 

I figured this PR is large enough as it is so best not add the AWS things yet.


internal/timeseries/out_streams.go
```diff
@@ -0,0 +1,71 @@
...
+		return PM
+	case aez.HTR:
+		return HTR
+	case aez.PWR:
+		return PWR
+	case aez.CPRU:
+		return CPRU
+	case aez.STAT:
+		return STAT
+	case aez.TCAcceptSuccessData, aez.TCAcceptFailureData, aez.TCExecSuccessData, aez.TCExecFailureData:
```
skymandr, line 66: In this clean implementation, this now looks very ugly to me... I wonder if I could have refactored it somehow?

skymandr (approved):
As far as I'm concerned this looks excellent. You may want wait for @joakimmoller to look at it as well though.

PR #99, opened 2020-05-19
Static compilation

### LINKED ISSUE #97: Build warning for statically linked lib
On `make build` the following warning is issued:
```rac-extract-payload> make build
go build -ldflags='-extldflags=-static' ./...
# github.com/innosat-mats/rac-extract-payload/cmd/rac
/tmp/go-link-173703299/000007.o: In function `_cgo_26061493d47f_C2func_getaddrinfo':
/tmp/go-build/cgo-gcc-prolog:58: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking```

joakimmoller:
Resolves #97 - VSC can build and test code.

Fixes some build/CI problems - "make lint" runs in CI in the same way as for users.

Golint version is specifically set in tools.go - i.e. `go mod tidy` can run without removing golint dependency.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,12 +16,6 @@ jobs:
         go-version: 1.14
       id: go
 
-    - name: setup env
-      run: |
-        echo "::set-env name=GOPATH::$(go env GOPATH)"
-        echo "::add-path::$(go env GOPATH)/bin"
-        go env
-
     - name: install mingw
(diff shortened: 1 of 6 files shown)


skymandr (approved):
Looks good to me and - more importantly - solves my problems. Even though @local-minimum didn't have exactly the same problems, I suggest you merge this since it's fairly critical. If his problems persist, we can reopen the issue.

PR #102, opened 2020-05-20
Pointerize

### LINKED ISSUE #83: Make pointer interface of Exporter
Current implementation causes copy of entire struct which for images holds MB of data, we should have `func (ccd *CCDImage) CSVRow()` (the asterisk) and so on instead

local-minimum:
# Content
Makes most fields in the DataRecord pointers and removes all methods like `STAT.Read(io.Reader)` and replaces them with `NewSTAT(io.Reader)`. This goes a bit further than the task specified but I figured it be quite confusing and unexpected that we have two patterns in the code (one for all in aez and another for common / innosat / ramses).

Running on 8 racs as a group this solution was actually 2% slower on my computer but it does have some advantages in the code (and some potential negative effects).

:heavy_check_mark: The switches that loads the various parts of aez in `instruments.go` is much nicer and very compact.
:heavy_check_mark: Though unlikely that some packet will get through misshaped to the callback, the CSV will keep format (specification used, number of header columns, number of data rows) and contrary to before it will be clear this has happened because missing data will be empty instead of 0.

:heavy_minus_sign: :question: Some more code on the csv functions of `DataRecord`.
:heavy_minus_sign: Some tests now need explicit `DataRecord` member initialization to avoid nil-pointer errors.

# Testing

Some tests got their expectations updated, but most were just alignments to being pointers.
Added tests for most new functions, but didn't include tests for the simple `NewXXX` functions that didn't have such before.

Resolves #83


internal/aez/htr.go
```diff
@@ -144,12 +146,12 @@ func (htr *HTR) Report() HTRReport {
 }
 
 //CSVHeaders returns the field names
-func (htr HTR) CSVHeaders() []string {
+func (htr *HTR) CSVHeaders() []string {
```
skymandr, line 149: What about the methods on lines 32--46? (Github won't let me comment up there.)

internal/aez/cpru.go
```diff
@@ -141,17 +143,17 @@ func (cpru *CPRU) Report() CPRUReport {
 }
 
 //CSVSpecifications returns the specs used in creating the struct
-func (cpru CPRU) CSVSpecifications() []string {
+func (cpru *CPRU) CSVSpecifications() []string {
```
skymandr, line 146: What about the methods on rows 16-54?

internal/aez/general.go
```diff
@@ -73,6 +73,11 @@ const (
 	PM RID = 30
 )
 
+// IsCCD returns if the RID is for a CCD
+func (rid RID) IsCCD() bool {
+	return rid == CCD1 || rid == CCD2 || rid == CCD3 || rid == CCD4 || rid == CCD5 || rid == CCD6 || rid == CCD7
+}
+
 func (rid RID) String() string {
```
skymandr: Why not pointers in these two?

internal/aez/htr.go
local-minimum: No really worth it since the type `htr` it's just a uint16... it should probably have another name as to not confuse it with the struct.

internal/aez/cpru.go
local-minimum: same argument.

internal/aez/pwr.go
```diff
@@ -164,17 +166,17 @@ func (pwr *PWR) Report() PWRReport {
 }
 
 //CSVSpecifications returns the specs used in creating the struct
-func (pwr PWR) CSVSpecifications() []string {
+func (pwr *PWR) CSVSpecifications() []string {
```
skymandr, line 178: Same comment/question here.

internal/aez/general.go
local-minimum: well I guess I can make it of *everything* but they're just scalar values and probably not copied as many times.

internal/innosat/source.go
```diff
@@ -99,25 +99,27 @@ type SourcePacketHeader struct {
 	PacketLength          uint16
 }
 
-//Read SourcePacketHeader
-func (sph *SourcePacketHeader) Read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, sph)
+// NewSourcePacketHeader reads a SourcePacketHeader from buffer
+func NewSourcePacketHeader(buf io.Reader) (*SourcePacketHeader, error) {
```
skymandr, line 103: ... and here ;)

internal/innosat/telemetry.go
```diff
@@ -26,9 +26,11 @@ type TMHeader struct {
 	CUCTimeFraction uint16
 }
 
-// Read TMHeader
-func (header *TMHeader) Read(buf io.Reader) error {
-	return binary.Read(buf, binary.BigEndian, header)
+// NewTMHeader reads a TMHeader from buffer
```
skymandr, line 29: and here

joakimmoller:
I'm net very concerned about the tiny slowdowns of execution speed. One can see from a trace run that GC kicks in more often that before but runs in shorter session (don't know if that is good or bad).

running "go test -race ./..." runs into problems though. But only the tests... the application runs fine.

skymandr (approved):
If @local-minimum and @joakimmoller are happy, then I'm happy. This makes the code a little harder to read, but I think it may be worth it from a didactic point of view.

PR #103, opened 2020-05-20
More info in README

local-minimum:
Adding some information about how no install and use.

Feel free to commit suggestions directly and merge when happy with it.


skymandr (approved):
Looks good to me.

PR #106, opened 2020-05-25
Including build info into application

### LINKED ISSUE #29: We need versioning of our code in output files!?
As per discussion with Donal, Ole Martin and Linda, my understanding is that produced artifacts of our code (time-series/house keeping data and images) need to be tagged with both what specification generated them as well as what version of our code.

I've tried to investigate this some and golang uses semantic versioning through tagging on e.g. GitHub and corresponding update in the `go.mod` (at the moment we are unversioned) file. As far as I can find after some googling, there's no way to access that information from within the binary. Which is a problem.

And technically, each time we modify the code in such a way that if we re-run a rac-file we don't get the exact same output, then we should update version.

How are we to solve this?

Perhaps: https://github.com/ofabry/version

local-minimum:
Adding a cli flag:
```
$ ~/go/bin/rac -version
Version v0.2.0 Commit e39b988 @ 2020-05-25T07:29:30Z
```

Adding info in the specifications row of the csv headers:

```
$ head -n 2 out/CCD.csv
CODE,v0.2.0 (e39b988) @ 2020-05-25T07:29:30Z,RAMSES,SPU045-S2:6F,INNOSAT,IS-OSE-ICD-0005:1,Specification,AEZICD002:H
File,ProcessingDate,RamsesTime,...
```

I tried to directly place the variables in `common.Version` etc. instead of `main.Version` but that doesn't seem to work and though a bit ugly perhaps this solution of `main` transferring its values to `common` is :ok: ?

@joakimmoller Is this working on windows?

The if-case in the `FullVersion` function is mostly there for testing. Building tests with ldflags didn't seem to work and besides this makes tests cases nicer.

**Limitation**: The version is just the most recent tag on the repository, so if we tag for other reason than making a release it will show that tag instead of the proper version.

Resolves #29

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 GOPATH ?= $(HOME)/go
 GOBIN ?= $(GOPATH)/bin
 GOCMD := go
-
+ifeq ($(OS),Windows_NT)
+SHELL := git-bash.exe
+endif
 # Fetch build info
 HEAD := $(shell git rev-parse --short HEAD)
 BUILDTIME := $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,24 @@
 GOPATH ?= $(HOME)/go
 GOBIN ?= $(GOPATH)/bin
 GOCMD := go
-GOBUILD := $(GOCMD) build
+
+# Fetch build info
+HEAD := $(shell git rev-parse --short HEAD)
+BUILDTIME := $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+VERSION := $(shell git describe --tags)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,13 +1,24 @@
 GOPATH ?= $(HOME)/go
 GOBIN ?= $(GOPATH)/bin
 GOCMD := go
-GOBUILD := $(GOCMD) build
+
```
joakimmoller: It works almost flawless (parsing date does not work) if you add:
```suggestion
ifeq ($(OS),Windows_NT)
SHELL := git-bash.exe
endif
```

joakimmoller (approved):
Works fine on windows except for parsing the date.

PR #107, opened 2020-05-25
Matrix CI

### LINKED ISSUE #91: CI needs to run on windows too

local-minimum:
Resolves #91 

* CI actually run on linux, windows and macOS.
* Fixed build artifacts too so they are built by make
* They can be found here https://github.com/innosat-mats/rac-extract-payload/actions/runs/114862892

@joakimmoller can you verify that the built rac.exe can both correctly display version and process rac files / generate image.

I opted for hardcoding the location of the binaries because the environment is rather obfuscated and the install process a bit flexible where it puts the files. Hope this is fine.

When PR is accepted I will update the `master` branch protection so that it requires the individual os CI to pass and remove the general 'Build' that is no more.


.github/workflows/go.yml
```diff
@@ -16,8 +19,9 @@ jobs:
         go-version: 1.14
       id: go
 
-    - name: install mingw
-      run : sudo apt-get install -y gcc-mingw-w64
+    - name: --Linux-- Install mingw
+      if: ${{ runner.os == 'Linux' }}
+      run: sudo apt-get install -y gcc-mingw-w64
```
joakimmoller: Don't think we need this now? This was to enable cross builds from Linux to Windows.

joakimmoller (approved):
Please remove mingw install. (It only runs for Linux?). Something is strange with the checks... ...waiting

.github/workflows/go.yml
local-minimum: Oh then I'll remove it. Reason it wasn't needed on windows is that the windows image comes with it pre-installed.

PR #108, opened 2020-05-25
Darwin compiles static

joakimmoller:
Static compilation after creating a symlink crt0.o -> crt1.o in the SDK


local-minimum (approved):
lgtm / I suppose you've tested it and it works

PR #110, opened 2020-05-26
Automating releases better

local-minimum:
* Uploads builds for all platforms
* ~Writes a release note~ Doesn't touch the release note so at least you can start writing it straight away.

I tried a million different combinations and pre-made actions but they don't do what we want, requires entirely different workflow, only works with npm-based repositories and a whole host of other issues. So I opt for the secondary objective, for the automation not to mess with what you've written so far and I have validated that it's now fine to start writing the release notes right away.

I think that we need to write our own action or write a script that collects the info we want and include running that script as a step. I did find one action where you can write your own manual text and probably that one can be used to include dynamic content from another step gathered into a file or environmental variable.

The text was written as the release was created in the UI.


.github/workflows/release.yml
```diff
@@ -9,39 +9,86 @@ jobs:
...
+      run: make install
+      shell: bash
+
+    - name: --Linux--- Bundle binaries
+      if: ${{ runner.os == 'Linux' }}
+      run: |
+        cd /home/runner/go/bin
+        tar -czvf "/home/runner/Rac_for_Linux.tar.gz" rac
+
+    - name: --macOS--- Bundle binaries
```
skymandr:
```suggestion
    - name: --Linux-- Bundle binaries
      if: ${{ runner.os == 'Linux' }}
      run: |
        cd /home/runner/go/bin
        tar -czvf "/home/runner/Rac_for_Linux.tar.gz" rac

    - name: --macOS-- Bundle binaries
```

skymandr (approved):
Looks very neat.

PR #118, opened 2020-06-02
CCD in name

### LINKED ISSUE #115: Add CCD number to filename
If the CCD 6 produced the image the filename should be changed:
```
MyArchive1_20200527-075857_20200527-080911_1589758163452.json ->
MyArchive1_20200527-075857_20200527-080911_1589758163452_6.json

MyArchive1_20200527-075857_20200527-080911_1589758163452.png ->
MyArchive1_20200527-075857_20200527-080911_1589758163452_6.png
```

local-minimum:
Resolves #115 

* Filename now looks like Myfile_50000000_3.png if it was from the RID CCD3
* Also fixes extra request from @lindamegner to add image file name into `CCD.csv`. Note that this will not include the project name, just the filename. It took some :thinking: to figure out how to do this. The project name isn't accessible when and where this is constructed and besides the `CCD.csv` will be in the same folder as the image referred to in it so it should be easy to find with just the filename.

I'll make a release once this is in master


internal/aez/ccd_image_test.go
```diff
@@ -116,3 +136,25 @@ func TestNewCCDImage(t *testing.T) {
...
+
+func TestCCDImage_FullImageName(t *testing.T) {
+	tests := []struct {
+		name          string
+		imageFileName string
+		prefix        string
+		want          string
+	}{
+		{"no prefix just filename", "test.png", "", "test.png"},
+		{"no prefix just filename", "😓️.png", "🌞️", filepath.Join("🌞️", "😓️.png")},
```
skymandr:
```suggestion
		{"filename with prefix", "😓️.png", "🌞️", filepath.Join("🌞️", "😓️.png")},
```

skymandr (approved):
Looks good to me, except one test that I think has the wrong description.

PR #132, opened 2021-06-22
Fixed a stats thing

### LINKED ISSUE #130: Better error statistics and reporting
It seems we have more orphaned packages than we should have sometimes. In order to help with debugging:

- [ ] Produce better error messages, so that we know why a package was lost
- [ ] Produce error statistics after interactive run: how many packages could be parsed out of how many total etc.

The research team might need some help looking into issues that are spotted during this work.

@OleMartinChristensen will provide test data for this. (Also, if you have any concrete suspicions or things we should have a look at, any additional information or similar, feel free to edit this issue!)

local-minimum:
For all three targets (stdout, disk, aws) we now get a final report.

Example usage:
```
$ rac -skip-images -skip-timeseries -stdout All\ Packets\ VC0_20210518-145413_20210519-094004.rac
Nothing will be extracted, only validating integrity of rac-file(s)

Statistics

Count    Error Message
116370   unhandled SID 2
43460    unhandled SID 4
36677    unhandled SID 3
24440    unhandled SID 5
12255    unhandled SID 7
12214    unhandled SID 6
8696     unexpected EOF
6146     unhandled SID 12
6146     unhandled SID 13
2021     orphaned multi-package data without termination detected
615      unhandled SID 9
408      unhandled SID 0
20       got continuation packet without a start packet
17       the TMHeader isn't recognized as either housekeeping, transparent or verification data (Service Type 5, Service Sub Type 1)
17       got stop packet without a start packet
5        the TMHeader isn't recognized as either housekeeping, transparent or verification data (Service Type 5, Service Sub Type 2)
1        the TMHeader isn't recognized as either housekeeping, transparent or verification data (Service Type 17, Service Sub Type 2)

Total Errors:	269508
Total Packages:	278014

```
Note that report omits the information about what package is the source of the error to be able to aggregate them better.

Resolves #130

### DIFF
--- a/internal/common/errorstats.go
+++ b/internal/common/errorstats.go
@@ -3,6 +3,7 @@ package common
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 )
 
@@ -28,28 +29,47 @@ func (stats *ErrorStats) Register(err error) {
 
 }
 
+func Max(x, y uint) uint {
+	if x < y {
+		return y
+	}
+	return x
+}
+
 // Summarize ...
 func (stats *ErrorStats) Summarize() string {
 	errs := make([]struct {
 		string
 		uint
 	}, 0)
+	if stats.errorsCount == 0 {
+		return fmt.Sprintf(
+			"\nStatistics\n\nTotal Errors:\t%v\nTotal Packages:\t%v\n",
+			stats.errorsCount,
+			stats.total,
+		)
+	}
+	var indent uint = 5 // "Count" is 5 characters
 	for key, value := range stats.errors {
+		indent = Max(indent, uint(len(strconv.Itoa(int(value)))))
 		errs = append(errs, struct {
 			string
 			uint
 		}{key, value})
 	}
+	indent += 3 // Spacing to next column
 	sort.SliceStable(errs, func(i, j int) bool {
 		return errs[i].uint > errs[j].uint
 	})
 	lines := make([]string, 0)
 	for _, item := range errs {
-		lines = append(lines, fmt.Sprintf("%v\t%s", item.uint, item.string))
+		lines = append(lines, fmt.Sprintf("%-*v%s", indent, item.uint, item.string))
 	}
 
 	return fmt.Sprintf(
-		"\nStatistics\n\nCount\tError Message\n%s\n\nTotal Errors:\t%v\nTotal Packages:\t%v\n",
+		"\nStatistics\n\n%-*vError Message\n%s\n\nTotal Errors:\t%v\nTotal Packages:\t%v\n",
+		indent,
+		"Count",
 		strings.Join(lines, "\n"),
 		stats.errorsCount,
 		stats.total,
--- a/internal/common/errorstats.go
+++ b/internal/common/errorstats.go
@@ -1,0 +1,57 @@
+package common
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// ErrorStats ...
+type ErrorStats struct {
+	total       uint // errors and not errors
+	errorsCount uint
+	errors      map[string]uint
+}
+
+// NewErrorStats ...
+func NewErrorStats() ErrorStats {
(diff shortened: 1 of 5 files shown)


internal/common/errorstats.go
```diff
@@ -0,0 +1,57 @@
...
+			string
+			uint
+		}{key, value})
+	}
+	sort.SliceStable(errs, func(i, j int) bool {
+		return errs[i].uint > errs[j].uint
+	})
+	lines := make([]string, 0)
+	for _, item := range errs {
+		lines = append(lines, fmt.Sprintf("%v\t%s", item.uint, item.string))
```
local-minimum: could make this a bit more complex if we want nice alignment of the second column probably.

internal/common/errorstats.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	return fmt.Sprintf(
+		"\nStatistics\n\nCount\tError Message\n%s\n\nTotal Errors:\t%v\nTotal Packages:\t%v\n",
```
local-minimum: could also make alternative statics for when there are no errors since it looks like this at the moment:
```

Statistics

Count	Error Message

Total Errors:	0
Total Packages:	278014

File: internal/common/errorstats.go (hunk 1 above)
Comment (skymandr) [reply internal/common/errorstats.go]: I think there are simple string formatting directives for padding, but the tab seems to do the trick well enough. If you think it is easy to make it more reliable then go ahead.

File: internal/common/errorstats.go (hunk 2 above)
Comment (skymandr) [reply internal/common/errorstats.go]: That would be good.

File: internal/exports/stdout_writer.go [hunk 3]
```diff
@@ -12,17 +12,21 @@ func StdoutCallbackFactory(
...
 
 	return func(pkg common.DataRecord) {
-		if writeTimeseries {
-			if pkg.Error != nil {
-				pkg.Error = fmt.Errorf(
-					"%s %s",
-					pkg.Error,
-					common.MakePackageInfo(&pkg),
-				)
+			errorStats.Register(pkg.Error)
```
skymandr, line 18: Looks like the indentation is off here, which is weird since that is not under your control, right? Please have a look at it nonetheless.

skymandr (changes requested):
Looks good, but at least fix the case when the errors are empty and have a look at the intent.
Making the table more reliable would be nice, but it seems good enough at the moment, so that is optional as far as I am concerned.

Don't forget to make a release when you are done.

internal/exports/stdout_writer.go
local-minimum: no go decides indentations and this is as it should be, though it looks dead ugly to me.

internal/common/errorstats.go
local-minimum: fixat

internal/common/errorstats.go
local-minimum: fixat

internal/common/errorstats.go
```diff
@@ -0,0 +1,77 @@
...
+			uint
+		}{key, value})
+	}
+	indent += 3 // Spacing to next column
+	sort.SliceStable(errs, func(i, j int) bool {
+		return errs[i].uint > errs[j].uint
+	})
+	lines := make([]string, 0)
+	for _, item := range errs {
+		lines = append(lines, fmt.Sprintf("%-*v%s", indent, item.uint, item.string))
```
skymandr, line 66: I think you should be able to use the `%<width>` syntax (e.g. `%42s`) rather than inserting an indent. This looks a bit clunky, but it's not important.

skymandr (approved):
I'm not super-happy with your solution, but so long as it works.

internal/common/errorstats.go
local-minimum: The `-` means left orient, if you do `%42` you get it right aligned, and the `*` means it gets the amount from an argument. So this is as it should be written according to the documentations.

PR #137, opened 2022-10-12
Implement correction for gpsTime

### LINKED ISSUE #135: UTC Time needs compensation
Same issue as in level1a/platform: https://github.com/innosat-mats/level1a-platform/issues/3, UTC Time needs 18 sec corr:

> The time in OHB files is time since GPS start - Obs according to IS-OSE-TN-0045_1 Reference Frames for InnoSat ACS.pdf there is a -18 second difference between GPS time and UTC. So the reference time used to compute date time needs to be reduced by 18 sec
> GPSstartime + datetime.timedelta(seconds = -18)
> 
> Found this when comparing GNSS positions with my own calculations

Clarification from @OleMartinChristensen:

>  In rac-extract the nanoseconds is extracted from two fields in the payload data package. This entry (nanoseconds) is correct. But you also convert this into a UTC date (which we have asked for), but in this conversion you assume that the epoch for the nanosecons is some other standard than the GPS time that the satellite uses (either you assume UTC of TIA, I'm not sure).
> 
> However we have checked and the discrepancy is 18 s. i.e. you should subtract 18s from the TMHeaderTime and EXP Date fields.

skymandr:
This subtracts 18 seconds from the `gpsTime` used as epoch.

Resolves #135


noreklint (approved):
LGTM, is the windows build expected to fail?

skymandr:
> LGTM, is the windows build expected to fail?

No, but it works now, after updating go-version.

PR #138, opened 2022-10-13
Upgrade aws

skymandr:
This upgrades the dependencies and, most crucially, updates to using v2 of the AWS cli. We should probably make a release once this and #135 are landed.


go.mod
```diff
@@ -3,13 +3,31 @@ module github.com/innosat-mats/rac-extract-payload
 go 1.19
 
 require (
-	github.com/aws/aws-sdk-go v1.31.0
+	github.com/aws/aws-sdk-go-v2 v1.16.16
+	github.com/aws/aws-sdk-go-v2/config v1.17.8
+	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.11.34
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.27.11
```
noreklint, line 9: I know in Typescript there's an issue if all internal cdk (not sure if it applies to sdk as well) dependencies aren't of the same version. These seem to be the latest version, so probably all good?

internal/awstools/uploader.go
```diff
@@ -1,28 +1,29 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/s3"
 )
 
 const awsBucket = "mats-l0-artifacts"
 
 // AWSUpload uploads file content to target bucket
-func AWSUpload(uploader *s3manager.Uploader, key string, body io.Reader) {
-	_, err := uploader.Upload(&s3manager.UploadInput{
+func AWSUpload(uploader *manager.Uploader, key string, body io.Reader) {
+	_, err := uploader.Upload(context.TODO(), &s3.PutObjectInput{
```
noreklint, line 18: What's this `context.TODO()`?

noreklint (changes requested):
LCTM, but some questions

go.mod
skymandr: Unfortunately (or fortunately? maybe the go version is more sensible than npm..?) the go packages don't seem to follow the versioning of the actual SDK. I've tried it locally and it works...

internal/awstools/uploader.go
skymandr: Surprised me too, but it seems to be what they do in the examples: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#NewUploader (see also: https://asanchez.dev/blog/amazon-s3-v2-golang/ for a more rigorous example).

The function is a go standard library function, documented here: https://pkg.go.dev/context#TODO.

go.mod
skymandr: (And when I say locally, I mean that I have processed a number of rac files locally with the `-aws` option and checked that they showed up in the expected bucket.)

noreklint (approved):

PR #139, opened 2022-10-19
Read RAC files from queue and process

### LINKED ISSUE #136: run rac on payload-batch
* place holder *

Run rac on data from queue. process  and put results on s3.

skymandr:
Resolves #136.

The lambda wakes up every 6 hours and tries to empty the notification queue for the bucket from `L0-fetcher`. Files are downloaded to local storage on the lambda and the rac binary is run on those.

Caveats:
- Not live tested (yet)!
- Not re-run friendly! A future improvement would be to wake up on a call from `L0-fetcher` with a list of what files to process. See #141
- Does not handle gaps from split multi-packets. See #142
- Does not write parquet, nor does it write to a prefix based on date, but that is the responsibility of the `rac` binary, I think. See issue #140


noreklint (approved):
:ok_hand:

PR #144, opened 2022-11-03
Make raclambda get triggered by l0fetcher events

### LINKED ISSUE #141: Make rac lambda listen to messages from L0-fetcher
The rac lambda currently wakes up periodically and empties the queue, but it would be better if it woke up on messages from https://github.com/innosat-mats/L0-fetcher containing a list of file-keys to download. This would make re-runs easier, as well as make the system at least slightly more streamlined.

# DoD
- [x] Fetcher lambda sends message
- [x] RAC lambda listens and parses
- [x]  Lambda för Platform data

noreklint:
The rac lambda will now listen to SQS events published by L0Fetcher. Each event triggers a single rac lambda and all files in that event message will be processed.

Resolves #141


skymandr (approved):
Much improved.

Issue #145, opened 2022-11-04
Retire S3 support from RAC binary


skymandr:
When processing in a lambda, the separation of concerns is a little shaky: the Lambda code downloads the data to local for the binary to process, but then the binary hands it off to a new bucket. This seems a little clumsy and complicated. We can already circumvent this by running the binary in local mode in the lambda, but in that case, do we need to keep the AWS support in the binary at all..?

# DoD
- [x] Remove AWS support in RAC code
- [x] `go mod tidy`
- [x] Update cli help and ReadMe
- [ ] ~Update Lambda if necessary~ This is better to wait with until #140

Issue #142, opened 2022-10-26
Handle gaps between rac batches


skymandr:
The rac binary and lambda work on batches of rac files, but sometimes an image will start in one batch and end in the next. Currently, these images are discarded. A proposed solution is to identify the dangling packages, write them to a temporary rac file which is stored somewhere predictable and downloaded and deleted/rewritten at the next batch processing. This would require that, or at least be made easier if, the `rac` binary associated image meta-data with the stop-packet rather than, as today, the start packet (the important piece of metadata is the file origin). See also #140.

PR #146, opened 2022-11-07
Handle gaps in RAC binary

skymandr:
This adds a new optional argument to the RAC binary which tells it where to look for and write "slask" files (if you don't like the name I can change it – suggestions welcome). Slask files are written with the TM Header time (CUC time) of the package whose buffer they represent as their filename.

When reading encountering a multi pack source packet without a start packet, the binary will look for the slask which matches the current stop or continuation packet according to the following criteria:
- the slask is _older_ than the current packet,
- the slask is not _too old_ (currently 30 seconds is allowed),
- the slask is the one _nearest in time_ given those criteria.
If an appropriate slask is found the binary will read it and concatenate with the current buffer, then processing proceeds as if nothing happened. Otherwise the binary will issue an error as before and continue with next packet.

This will necessitate a new release, but I think it should be a minor release (when we start using parquet and/or remove AWS support as in #145 , maybe we should go for major).

Resolves the RAC part of #142.

Remains:
- [ ] RAC lambda should read slask from bucket
- [ ] RAC lambda should give binary information on where to find them
- [ ] RAC lambda should upload new slask files to bucket


noreklint (approved):
If I were to look at this repository without any prior information about it I would probably think "What on earth (or in space) is this slask thing? 
I think maybe consider renaming it or just add a brief explanation in a docstring at the `Slask` struct. I'll approve anyway so if you don't agree just skip it

skymandr:
> If I were to look at this repository without any prior information about it I would probably think "What on earth (or in space) is this slask thing? I think maybe consider renaming it or just add a brief explanation in a docstring at the `Slask` struct. I'll approve anyway so if you don't agree just skip it

I'd be happy to change the name, but I don't know to what! Any suggestions?

PR #147, opened 2022-11-09
Implement support for continuing multi packets in lambda

### LINKED ISSUE #142: Handle gaps between rac batches

skymandr:
This adds support for continuing multi packets in the rac lambda.

Continuation files ("dregs") are downloaded from a purpose made bucket with a retention policy of 7 days, created in this stack.
Files are saved locally in the lambda and read as needed.
At the end of the process, the lambda checks if there are any new continuation files and if so uploads them to the bucket.

Resolves #142


raclambda/raclambda/handler/raclambda_handler.py
```diff
@@ -1,43 +1,122 @@
...
+        ])
+
+        # Upload new Dregs
+        new_dregs = get_new_files(dregs_dir, dregs)
+        upload_files(s3_client, dregs_bucket, dregs_dir, new_dregs)
 
 
 if __name__ == "__main__":
-    handler(None, None)
+    handler({}, None)
```
skymandr, line 122: I don't think this `if` is needed. I'm feeling inclined toward remove it.

raclambda/raclambda/handler/raclambda_handler.py
```diff
@@ -1,43 +1,122 @@
...
 
 
-def handler(event, context):
-    project = get_env_or_raise("RAC_PROJECT")
-    rac_dir = "/tmp"
+def get_all_files(
+    bucket_name: str,
+) -> List[str]:
+    bucket = boto3.resource("s3").Bucket(bucket_name)
+    return [obj.key for obj in bucket.objects.all()]
```
skymandr, line 37: I didn't think there was any point in testing this.

raclambda/raclambda/raclambda_stack.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         project_name: str,
         queue_arn_export_name: str,
         lambda_timeout: Duration = Duration.seconds(300),
+        dregs_expiration: Duration = Duration.days(7),
```
noreklint, line 29: This means that files will only be kept for a week, right?
Is this the mechanism that cleans up used dregs that are no longer useful?

noreklint (approved):
LGTM but some questions about the DREGS bucket handling

raclambda/raclambda/raclambda_stack.py
skymandr: Yes, that is the idea.

PR #148, opened 2022-11-10
Remove AWS support in RAC binary

### LINKED ISSUE #145: Retire S3 support from RAC binary

skymandr:
This removes the AWS support from the RAC binary.

Resolves #145.


noreklint (approved):
:wastebasket: :+1:

PR #149, opened 2022-11-21
Write paRACquet

### LINKED ISSUE #140: Output RAC data as parquet
The rac binary should be able to output parquet files, one per rac file. At least when writing to AWS it might be beneficial for the processing to postfix the project prefix with y/m/d. For the benefit of future dealing with gaps, it would be good if images from multi-packs were associated with the meta-data of the stop packet, rather than the start packet (the important piece of meta data is the file origin; see also #142).

skymandr:
This resolves #140, though the lambda needs updating as well.

If the "-parquet" flag is used, data will now be written to a structure like "project/year/month/day". One parquet file is written per rac file per partition it covers. Typical parquet-sizes are ~25 MB, which is on the small side, and they may sometimes be much smaller. Typically each partition seems to hold 10–20 of those.

Capability to write to CSV is retained as default. This writes the same files as before, but some headers etc. have been updated in the resulting CSVs.


README.md
```diff
@@ -12,6 +12,8 @@ Run the binary (if you are on windows `rac.exe`):
 
 The `-project` sets output directory in this case.
 
+The `-stdout` print output instead of writing to disk, ignoring images.
```
noreklint, line 15: This should probably include info about `-parquet` option

internal/timeseries/parquet_collection.go
```diff
@@ -0,0 +1,73 @@
...
+
+// ParquetName returns the whole name of the parquet, including partitioning prefix
+func ParquetName(pkg *common.DataRecord) string {
+	tmTime := pkg.TMHeader.Time(time.Time{})
+	prefix := filepath.Join(
+		fmt.Sprintf("%v", tmTime.Year()),
+		fmt.Sprintf("%v", int(tmTime.Month())),
+		fmt.Sprintf("%v", tmTime.Day()),
+	)
+	baseName := filepath.Base(pkg.Origin.Name)
```
noreklint, line 35: Is `pkg.Origin.Name` the `.rac` file name? If not, can it be the same for multiple `pkg`?

noreklint (changes requested):
Looking good, just some small remarks

internal/timeseries/parquet_collection.go
skymandr: It is the name of the rac-file.

README.md
skymandr: Fixed.

noreklint (approved):

PR #151, opened 2022-11-24
Write different schemas

skymandr:
This changes so that parquet files are written to different files based on both packet type and original file, using different parquet schemas. This is preferable, since otherwise golang will helpfully add default values (0) to columns that belong to another packet type, making rows hard to disambiguate. The new file naming scheme also makes it easy to filter what files to read in PyArrow, which might be useful.


noreklint (approved):
Smooth sailing :boat:

PR #152, opened 2022-11-28
Updated lambda to output parquet and upload using rclone

skymandr:
What it says on the box: Lambda now uses updated rac binary to output parquet. The output is copied to the output bucket by the lambda using rclone. In addition, copying  to and from the dregs bucket is now also performed using rclone.

Final part of #140


noreklint (approved):
:+1:

PR #155, opened 2022-11-30
Fix error handling in decode jpeg

### LINKED ISSUE #153: Problem using rac extract on mulitple files
# rac extract v 2.7

This was investigated sorting datetimes and looking at first and last items.

Racfile download dates
EXP Dates in files

21st to 28th: 
EXP Date from 22-11-22 08:32:54 to 22-11-23 19.16:07:25.

25th to 28th:
EXP Date from 2022-11-24 00:01:20 to 2022-11-24 00:05:31.

26th to 28th:
EXP Date from 2022-11-24 18:06:05 to 2022-11-24 19:00:03.

Something wrong when reading in multiple files?

# To reproduce
Download all .rac data from OHB FTP server and try and run rac extract on all of them.

skymandr:
This fixes the error handling in the JPEG decoding, which resolves #153.

The problem was that LibJPEG:s default error handler has some errors where it issues an `exit()`. The implementer using the library has to override this, if the library is used embedded in another application, see _e.g._ https://github.com/LuaDist/libjpeg/blob/master/example.c

This also fixes at least some error propagation from the decoding to the rest of the code. The method we used before didn't work. This at least propagates the most sever errors, but a lot of other errors (or warnings) issued by LibJPEG can be seen when running `rac` on one of the corrupt files. These other errors are printed by LibJPEG itself, and are currently not handled at all in our code, but they don't cause a crash.

We could handle them like the more severe errors, but the only thing we do with the errors from decoding at the moment is to log them, which LibJPEG sort of already does, so this seems to be of limited use.

### DIFF
--- a/internal/decodejpeg/decode.go
+++ b/internal/decodejpeg/decode.go
@@ -23,10 +23,7 @@ func JpegImageData(jpegData []byte) (rawData []uint16, height int, width int, er
 	jpegChar := C.CString(string(jpegData))
 	defer C.free(unsafe.Pointer(jpegChar))
 
-	jErrBuf := ""
-	for i := 0; i < C.JMSG_LENGTH_MAX; i++ {
-		jErrBuf += " "
-	}
+	jErrBuf := strings.Repeat(" ", CJMSG_LENGTH_MAX)
 
 	jpegErr := C.CString(string(jErrBuf))
 	defer C.free(unsafe.Pointer(jpegErr))
--- a/internal/decodejpeg/decode.go
+++ b/internal/decodejpeg/decode.go
@@ -23,7 +23,7 @@ func JpegImageData(jpegData []byte) (rawData []uint16, height int, width int, er
 	jpegChar := C.CString(string(jpegData))
 	defer C.free(unsafe.Pointer(jpegChar))
 
-	jErrBuf := strings.Repeat(" ", CJMSG_LENGTH_MAX)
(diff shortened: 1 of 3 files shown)


internal/decodejpeg/decode.go
```diff
@@ -12,22 +12,38 @@ import "C"
...
 	jpegChar := C.CString(string(jpegData))
 	defer C.free(unsafe.Pointer(jpegChar))
 
-	imageData, err := C.read_JPEG_file(jpegChar, C.size_t(len(jpegData)))
-	if err != nil {
-		return rawData, height, width, err
+	jErrBuf := ""
+	for i := 0; i < C.JMSG_LENGTH_MAX; i++ {
+		jErrBuf += " "
 	}
```
joakimmoller:
```suggestion
	jErrBuf := strings.Repeat(" ", CJMSG_LENGTH_MAX)
```

joakimmoller (changes requested):
What happens when a image could not be decoded, will the rest of the data in the CCD-struct also be disregarded?

joakimmoller (approved):

internal/decodejpeg/decode.go
skymandr: Thanks! That annoyed me to no end...

skymandr:
> What happens when a image could not be decoded, will the rest of the data in the CCD-struct also be disregarded?

No, the program is designed to save data regardless.

PR #159, opened 2022-12-02
Deploy

noreklint:
Pretty much same story as the other services

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -3,6 +3,10 @@ name: Rac Lambda CD
 on:
   workflow_dispatch:
 
+defaults:
+  run:
+    working-directory: raclambda
+
 jobs:
   aws_cdk:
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -0,0 +1,45 @@
+name: Rac Lambda CD
+
+on:
+  release:
+    types: [released]
```
skymandr, line 5: Doesn't this give us (a blunt, but still) tool for controlling deployment, by making _e.g._ legacy-releases pre-releases only?

skymandr (approved):
LGTM. Note that ephemeral disk space is currently not set explicitly in stack, which is likely to be an issue. It should probably be as large as the memory.

.github/workflows/cd.yml
noreklint: Yep, that's true

PR #160, opened 2022-12-02
Add tweaks to resolve deployment issues

noreklint:
- Fixed a problem where the lambda couldn't find `./rac` because of wrong working directory
- Fixed a problem where the rac files couldn't be found, it was looking for the file `/path/*.rac` rather than the result of the wildcard query
- Removed `project` from the lambda since it didn't fit with the scheme of buckets describing data version

### DIFF
--- a/raclambda/tests/raclambda/handler/test_raclambda_handler.py
+++ b/raclambda/tests/raclambda/handler/test_raclambda_handler.py
@@ -74,7 +74,7 @@ def test_handler(monkeypatch):
         'raclambda.handler.raclambda_handler.download_files',
     ) as patched_download, patch(
         'raclambda.handler.raclambda_handler.Path.glob',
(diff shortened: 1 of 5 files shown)


raclambda/tests/raclambda/handler/test_raclambda_handler.py
```diff
@@ -73,6 +73,9 @@ def test_handler(monkeypatch):
     ) as patched_rclone_config, patch(
         'raclambda.handler.raclambda_handler.download_files',
     ) as patched_download, patch(
+        'raclambda.handler.raclambda_handler.Path.glob',
+        return_value=rac_files
```
skymandr:
```suggestion
        return_value=rac_files,
```

skymandr (approved):
Looks cool to me. 🕶️

PR #162, opened 2022-12-06
Add stream type prefix before partitioning rather than to filename

skymandr:
This changes the output structure from 

`y/m/d/STREAM_filename.parquet`

to

`STREAM/y/m/d/filename.parquet`

This structure is preferable, since it makes it very easy for the lambda listening to output from this one to know if it should wake up or not. The downside is that this makes it a little less convenient to read data from different sources at the same time, but no more so, than from different CSVs, for instance.

I suggest this is a minor-release-worthy change, so bump to 1.2.0 after this.


noreklint (approved):
:+1:

PR #167, opened 2022-12-14
Refactor lambda to expect and handle a single file at a time

noreklint:
Part of https://github.com/innosat-mats/level1a-platform/issues/14

It now expects notifications to include an `object` rather than multiple `objects`, and handles that one accordingly


skymandr (approved):
I think you should lower the memory and ephemeral storage. The largest rac-files we've seen are smaller than 200 MB. We need twice that in storage and two to four times that in memory.

Other than that, looks good!

Issue #170, opened 2023-01-16
How to re-run?


skymandr:
We would like to have an easy way of re-running from scratch. For most services, this can be done easily by a manual copy to self in the input bucket using, but for this service the queue is FIFO. Perhaps a script taking a date range or start-date as input?

#### Consider if this should be solved in L0-fetcher and make it usable for both platform and rac

See https://stackoverflow.com/questions/30249069/listing-contents-of-a-bucket-with-boto3

skymandr:
See https://stackoverflow.com/questions/30249069/listing-contents-of-a-bucket-with-boto3

PR #174, opened 2023-01-31
Partition by hour

### LINKED ISSUE #169: Partition by hour?
When deciding partitioning scheme we thought we would be getting several rac-files each day, but we seem to only be geting one. Therefore, partitioning by hour might be feasible.

By partitioning by hour, we might avoid hick-ups along the line, have an easier time debugging them when they occur, and it will be easier to get to a point where we can produce L1B efficiently.

skymandr:
Resolves #169

Before deploying (which is done automatically when a new tag is created, but not when merging) buckets should be created the following steps in the chain coordinated with new buckets both for input and output.

It would also be nice if #170 was finished before deploying, so that we can make a quick re-run.

### DIFF
--- a/internal/exports/parquet_writer_test.go
+++ b/internal/exports/parquet_writer_test.go
@@ -34,8 +34,8 @@ func Test_parquetName(t *testing.T) {
 		args args
 		want string
 	}{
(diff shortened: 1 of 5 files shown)


noreklint (approved):
LGTM

